PR #9, opened 2019-11-28
[TASK] Convert the PHP namespaces to "TTN"

### LINKED ISSUE #1: Switch the PHP namespaces to the TTN

oliverklee:
Fixes #1

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -15,6 +15,7 @@ This project adheres to [Semantic Versioning](https://semver.org/).
 - Composer script for PHP linting
 
 ### Changed
+- Convert the PHP namespaces to "TTN" (#8)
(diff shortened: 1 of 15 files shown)


oliverklee:
@andreaswolf @DanielSiepmann Could you please test whether the functional controller tests run for you? I get failures as some part of the code still is using the old namespace, and I have no idea why.

andreaswolf:
@oliverklee Try throwing away the contents of `public/typo3temp/var/tests/`. There are things that are copied over there for functional tests, maybe that affects the tests.

oliverklee:
@andreaswolf Thanks for the hint. `.Build/public/typo3temp/var/tests/` wasn't the culprit, but `var/cache/` was. Thanks again!

PR #23, opened 2019-12-07
Fix codesniffer issues

andreaswolf:
Resolves: #21

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -6,6 +6,7 @@ This project adheres to [Semantic Versioning](https://semver.org/).
 ## x.y.z
 
 ### Added
+- Composer script for PHP code sniffer fixing (#21)
 - Run the functional tests in parallel
 - Add PHP-CS-Fixer
 - Add support for PHP 7.3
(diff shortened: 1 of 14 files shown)


composer.json
```diff
@@ -65,6 +65,7 @@
     "scripts": {
         "ci:php:lint": "find *.php Classes/ Configuration/ Tests/ -name '*.php' -print0 | xargs -0 -n 1 -P 4 php -l",
         "ci:php:sniff": "phpcs Classes Tests",
+        "ci:php:fix": "phpcbf Classes Tests",
```
oliverklee: This should not be in the `ci` namespace as this isn't something to be run during CI, but during development. I propose `fix:php-sniff` (as there might also be another fixer using PHP_CS_Fixer later).

CHANGELOG.md
```diff
@@ -13,6 +13,7 @@ This project adheres to [Semantic Versioning](https://semver.org/).
 - Add PHP_CodeSniffer to the Travis CI build
 - Auto-release to the TER
 - Composer script for PHP linting
+- Composer script for PHP code sniffer fixing
```
oliverklee: Please add the latest line on top, and also please add the PR number.

PR #28, opened 2019-12-07
[TASK] Move PHP linting to GitHub actions

andreaswolf:
Resolves: #12

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -1,0 +1,25 @@
+on:
+  - pull_request
+  - push
+
+name: CI
+
+jobs:
+  php-lint:
+    name: PHP linter
+
+    runs-on: ubuntu-latest
+
+    strategy:
+      matrix:
+        php-version:
+          - 7.2
+          - 7.3
+          - 7.4
+
+    steps:
+    - name: Checkout
+      uses: actions/checkout@v1
+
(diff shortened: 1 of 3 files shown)


.github/workflows/ci.yml
```diff
@@ -0,0 +1,25 @@
...
+          - 7.2
+          - 7.3
+          - 7.4
+
+    steps:
+    - name: Checkout
+      uses: actions/checkout@v1
+
+    - name: Run PHP lint
+      run: composer run-script ci:php:lint
```
oliverklee: Why not the shorter form?

`composer ci:php:lint`

CHANGELOG.md
```diff
@@ -16,6 +16,7 @@ This project adheres to [Semantic Versioning](https://semver.org/).
 - Composer script for PHP linting
 
 ### Changed
+- Moved PHP linting to GitHub actions (#12)
```
oliverklee: … and also #28

andreaswolf: I don't see why we need PR + issue ID, esp. since the PR ID can't be known beforehand.

.github/workflows/ci.yml
andreaswolf: true… I think that's a leftover from the GitHub action template I had a look at.

PR #30, opened 2019-12-07
[TASK] Move PHP code sniffing to GitHub actions

andreaswolf:
Resolves: #13

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -23,3 +23,18 @@ jobs:
 
     - name: Run PHP lint
       run: composer ci:php:lint
+
+  php-code-sniffer:
+    name: PHP Code Sniffer
+
+    runs-on: ubuntu-latest
+
+    steps:
+    - name: Checkout
+      uses: actions/checkout@v1
+
+    - name: Install Composer dependencies
+      run: composer install --no-progress
+
+    - name: Run PHP Code Sniffer
+      run: composer ci:php:sniff
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -16,7 +16,8 @@ This project adheres to [Semantic Versioning](https://semver.org/).
 - Composer script for PHP linting
 
 ### Changed
-- Moved PHP linting to GitHub actions (#12)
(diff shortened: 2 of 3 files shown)


.github/workflows/ci.yml
```diff
@@ -23,3 +23,25 @@ jobs:
...
+          - 7.2
+          - 7.3
+          - 7.4
+
+    steps:
+    - name: Checkout
+      uses: actions/checkout@v1
+
+    - name: Install Composer dependencies
+      run: composer install --prefer-dist --no-progress
```
oliverklee: The prefer-dist already is configured in the `composer.json`:
https://github.com/typo3-trainer-network/tea/blob/master/composer.json#L59

CHANGELOG.md
```diff
@@ -16,6 +16,7 @@ This project adheres to [Semantic Versioning](https://semver.org/).
 - Composer script for PHP linting
 
 ### Changed
+- Moved PHP code sniffing to GitHub actions (#13)
```
oliverklee: In general: Moved -> Move

.github/workflows/ci.yml
```diff
@@ -23,3 +23,25 @@ jobs:
...
+    name: PHP Code Sniffer
+
+    runs-on: ubuntu-latest
+
+    strategy:
+      matrix:
+        php-version:
+          - 7.2
+          - 7.3
+          - 7.4
```
oliverklee: For sniffing, one PHP version should suffice, as the result will not depend on the PHP version running the sniffer.

CHANGELOG.md
andreaswolf: I tend to disagree, since the headline is also past tense. But then we should also fix all existing entries.

oliverklee: I usually copy the Git commit message headlines into the changelog, resulting in this form.
https://chris.beams.io/posts/git-commit/

(But I'm not very passionate about the tenses in the changelog.)

andreaswolf: Sounds logical, let's stick to present tense then.

PR #38, opened 2019-12-07
[TASK] Move unit test run to GitHub actions

andreaswolf:
Resolves: #37

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -77,6 +77,9 @@ jobs:
 
     strategy:
       matrix:
+        typo3-version:
+          - "^9.5"
+          #- "^10.2"
         php-version:
           - 7.2
           - 7.3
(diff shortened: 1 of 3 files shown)


.github/workflows/ci.yml
```diff
@@ -69,3 +69,33 @@ jobs:
...
+    name: Unit tests
+
+    runs-on: ubuntu-latest
+
+    strategy:
+      matrix:
+        php-version:
+          - 7.2
+          - 7.3
+          - 7.4
```
oliverklee: We'll also need to run the tests with TYPO3 9 and 10.

andreaswolf: This is currently not possible since the dev-dependencies do not work with TYPO3 v10.

.github/workflows/ci.yml
```diff
@@ -69,3 +69,43 @@ jobs:
...
+  unit-tests:
+    name: Unit tests
+
+    runs-on: ubuntu-latest
+
+    strategy:
+      matrix:
+        typo3-version:
+          - "^9.5"
+          #- "^10.2"
```
oliverklee: Is this intentional?

PR #39, opened 2019-12-07
[BUGFIX] Keep the global namespace clean in ext_localconf.php

oliverklee:
Move the plugin registration into a function.


ext_localconf.php
```diff
@@ -1,16 +1,23 @@
...
+        ],
+        // non-cacheable actions
+        [
+            'Tea' => '',
+        ]
+    );
+};
+
+$boot();
+unset($boot);
```
andreaswolf: IMO this could also be `(static function () {…})();`

oliverklee: Changed and repushed.

PR #48, opened 2020-05-21
[BUGFIX] Use the PHP version from the matrix in the CI

oliverklee:
For GitHub actions, an explicit step is required to use a
specific PHP version. If this is not done, a default version
will be used (which makes the builds unpredictable).


oliverklee:
I have rebase. I'll merge this PR in a monite.

PR #49, opened 2020-05-21
[BUGFIX] Always use the Composer-installed tools

oliverklee:
Add the full relative path to Composer-installed binaries to make
sure that are used instead of other versions that happen to be in
the path.

Also drop unneeded trailing slashes from directories used in
Composer scripts.

### DIFF
--- a/composer.json
+++ b/composer.json
@@ -63,11 +63,11 @@
         "vendor-dir": ".Build/vendor"
     },
     "scripts": {
-        "ci:php:lint": "find *.php Classes/ Configuration/ Tests/ -name '*.php' -print0 | xargs -0 -n 1 -P 4 php -l",
-        "ci:php:sniff": "phpcs Classes Tests",
-        "ci:ts:lint": "typoscript-lint -c Configuration/TsLint.yml --ansi -n --fail-on-warnings -vvv Configuration/TypoScript/",
-        "ci:tests:unit": "phpunit -c .Build/vendor/nimut/testing-framework/res/Configuration/UnitTests.xml Tests/Unit/",
-        "ci:tests:functional": "find 'Tests/Functional' -wholename '*Test.php' | parallel --gnu 'echo; echo \"Running functional test suite {}\"; phpunit -c .Build/vendor/nimut/testing-framework/res/Configuration/FunctionalTests.xml {}';",
+        "ci:php:lint": "find *.php Classes Configuration Tests -name '*.php' -print0 | xargs -0 -n 1 -P 4 php -l",
+        "ci:php:sniff": ".Build/vendor/bin/phpcs Classes Tests",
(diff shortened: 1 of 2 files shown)


composer.json
```diff
@@ -63,11 +63,11 @@
         "vendor-dir": ".Build/vendor"
     },
     "scripts": {
-        "ci:php:lint": "find *.php Classes/ Configuration/ Tests/ -name '*.php' -print0 | xargs -0 -n 1 -P 4 php -l",
-        "ci:php:sniff": "phpcs Classes Tests",
-        "ci:ts:lint": "typoscript-lint -c Configuration/TsLint.yml --ansi -n --fail-on-warnings -vvv Configuration/TypoScript/",
-        "ci:tests:unit": "phpunit -c .Build/vendor/nimut/testing-framework/res/Configuration/UnitTests.xml Tests/Unit/",
-        "ci:tests:functional": "find 'Tests/Functional' -wholename '*Test.php' | parallel --gnu 'echo; echo \"Running functional test suite {}\"; phpunit -c .Build/vendor/nimut/testing-framework/res/Configuration/FunctionalTests.xml {}';",
+        "ci:php:lint": "find *.php Classes Configuration Tests -name '*.php' -print0 | xargs -0 -n 1 -P 4 php -l",
+        "ci:php:sniff": ".Build/vendor/bin/phpcs Classes Tests",
```
DanielSiepmann, line 67: I understood that composer will add the `bin` folder to the beginning of current `$PATH`, so this should not be necessary.
Is this information wrong?

oliverklee: Yes, at least for the Composer provided by `shivammathur/setup-php@v2` for GitHub actions (or maybe also for GitHub actions in general). In https://github.com/MyIntervals/emogrifier, we had exactly the problem that the system-installed tools were called instead of the tools from the current project when we switched from TravisCi to GitHub Actions.

(On my local machine, I added `./vendor/bin/`to my path manually.)

I could not find a reference saying that Composer automatically modifies the current `$PATH`. Do you have a reference for this?

DanielSiepmann: > Note: Before executing scripts, Composer's bin-dir is temporarily pushed on top of the PATH environment variable so that binaries of dependencies are easily accessible. In this example no matter if the phpunit binary is actually in vendor/bin/phpunit or bin/phpunit it will be found and executed.
> — https://getcomposer.org/doc/articles/scripts.md#writing-custom-commands

Maybe that's not true for some GitHub Actions. In that case I would be fine with merging.
But Yarn, NPM and Composer should all work this way.

Some projects don't use the default `./vendor/bin` but `./.Build/bin`, `./Build/bin` or `./bin` or some other path. E.g. TYPO3 itself uses `./bin`.

PR #52, opened 2020-05-22
[BUGFIX] Stop caching vendor/ on Travis CI

oliverklee:
Downgrades of Composer packages do not work reliably
if a higher version of a package already is installed in `vendor/`.
This causes builds to use higher versions of packages than intended
if Travis CI has the corresponding version in the cached and restored
`vendor/` folder.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -16,7 +16,6 @@ env:
 
 cache:
   directories:
-  - .Build/vendor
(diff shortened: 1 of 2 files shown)


DanielSiepmann:
As far as I know, this will be fixed with release of composer 2.x.

PR #55, opened 2020-05-27
[FEATURE] Run the functional tests via GitHub Actions

oliverklee:


.github/workflows/ci.yml
```diff
@@ -151,3 +151,68 @@ jobs:
...
+  functional-tests:
+    name: Functional tests
+
+    runs-on: ubuntu-latest
+
+    strategy:
+      matrix:
+        typo3-version:
+          - "^9.5"
+          #- "^10.2"
```
DanielSiepmann, line 164: I guess `10.4` instead of `10.2`.

oliverklee: Done and repushed.

PR #59, opened 2020-06-22
[TASK] Add yaml lint

lukaszuznanski:
Add yaml lint and lint all files.


.github/workflows/ci.yml
```diff
@@ -1,218 +1,191 @@
...
-      run: composer ci:php:lint
-
+      - name: Checkout
+        uses: actions/checkout@v1
+      - name: Install PHP
+        uses: shivammathur/setup-php@v2
+        with:
+          php-version: '${{ matrix.php-version }}'
+      - name: Run PHP lint
+        run: 'composer ci:php:lint'
```
oliverklee, line 24: @lukaszuznanski I find this PR quite hard to read due to the mix of formatting changes and other "real" changes. Would you mind creating a PR for the formatting changes that we can merge first? That would be very helpyful for me. Thanks!

.travis.yml
```diff
@@ -1,66 +1,52 @@
+---
```
oliverklee, line 1: We've just removed Travis CI support in #56. So this file does not exist anymore.

.github/workflows/ci.yml
oliverklee: (For the other files, this is fine as there are only formatting changes as far as I can see. Only this file has the mix of formatting changes and new/changed code.)

lukaszuznanski: I get your point. I will separate those changes.

.travis.yml
lukaszuznanski: Ok, fine, I will remove this file

.github/workflows/ci.yml
oliverklee: Thanks!

lukaszuznanski:
requires #72 merge first

lukaszuznanski:
If it's fine, we should merge this PR as soon as possible, because making further changes to ci.yml would be much easier :)

PR #78, opened 2020-07-22
[FEATURE] Add security check

lukaszuznanski:


.github/workflows/ci.yml
```diff
@@ -197,3 +197,28 @@ jobs:
...
+          path: ~/.composer/cache
+          restore-keys: "php${{ matrix.php-version }}-composer-\n"
+      - name: "Install Composer dependencies"
+        run: "composer install --no-progress"
+      - name: "Run PHP Security test"
+        run: "composer ci:tests:security"
+    strategy:
+      matrix:
+        php-version:
+          - 7.4
```
oliverklee, line 224: Please let's move the matrix up so the reader has seen it when they see the matrix variable getting used.

composer.json
```diff
@@ -72,9 +73,11 @@
         "ci:ts:lint": ".Build/vendor/bin/typoscript-lint -c Configuration/TsLint.yml --ansi -n --fail-on-warnings -vvv Configuration/TypoScript",
         "ci:tests:unit": ".Build/vendor/bin/phpunit -c .Build/vendor/nimut/testing-framework/res/Configuration/UnitTests.xml Tests/Unit",
         "ci:tests:functional": "find 'Tests/Functional' -wholename '*Test.php' | parallel --gnu 'echo; echo \"Running functional test suite {}\"; .Build/vendor/bin/phpunit -c .Build/vendor/nimut/testing-framework/res/Configuration/FunctionalTests.xml {}';",
+        "ci:tests:security": ".Build/vendor/bin/security-checker security:check",
         "ci:tests": [
             "@ci:tests:unit",
-            "@ci:tests:functional"
+            "@ci:tests:functional",
+            "@ci:tests:security"
```
oliverklee, line 80: Are those really tests in the meaning of "dynamic tests along the lines of unit tests, functional tests, acceptance tests"? If not, I'd opt for moving this from `ci:tests:security` to `ci:security`.

.github/workflows/ci.yml
lukaszuznanski: You mean before "steps"? If so, then we should also move this for other jobs

oliverklee: Yes, before "steps". For the existing jobs, please let's do this in a separate PR so we don't mix things up.

PR #81, opened 2020-07-24
[TASK] Add editorconfig

lukaszuznanski:


.editorconfig
```diff
@@ -0,0 +1,65 @@
...
+
+# YML-Files
+[{*.yml,*.yaml}]
+indent_style = space
+indent_size = 2
+
+# JSON-Files
+[{*.jsonl}]
+indent_style = space
+indent_size = 2
```
oliverklee, line 65: Should we have a trailing newline after the last line?

Issue #83, opened 2020-07-24
Reduce the GitHub actions builds


oliverklee:
https://docs.github.com/en/actions/reference/events-that-trigger-workflows#example-using-multiple-events-with-activity-types-or-configuration

Maybe we can reduce executing the same action twice.

lukaszuznanski:
Right now we are running on every single action, so when I push some code and create PR, then we have build, after merge to master there is another build. You are saying that we should disable for example merge to master build?

PR #86, opened 2020-07-24
[TASK] Reduce GitHub actions builds

lukaszuznanski:
Resolves: #83

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -88,7 +88,6 @@ jobs:
   unit-tests:
     name: "Unit tests"
     runs-on: ubuntu-latest
-    needs: php-lint
     steps:
       -
         name: Checkout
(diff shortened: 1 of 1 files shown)


.github/workflows/ci.yml
```diff
@@ -1,8 +1,12 @@
...
 name: CI
 on:
-  - pull_request
-  - push
+  push:
+    branches:
+      - master
+  pull_request:
+    branches:
+      - master
```
oliverklee, line 9: Could we also use only "push" (for all branches)? Then we'd cover new PRs (which have been created from a recently pushed branch) as well as pushed to existing PRs and merges to master.

.github/workflows/ci.yml
```diff
@@ -1,8 +1,10 @@
 ---
 name: CI
 on:
-  - pull_request
-  - push
+  push:
+  pull_request:
```
oliverklee, line 5: Now we still have the `pull_request` trigger. Can we reduce this to only having the `push`trigger?

lukaszuznanski:
I will take care about this PR after call :)

PR #87, opened 2020-08-04
[TASK] Add gitlab autochecks

lukaszuznanski:


lukaszuznanski:
https://gitlab.typo3.org/qa/example-extension/-/pipelines/9368

.gitlab-ci.yml
```diff
@@ -23,3 +29,15 @@ build-composer-dependencies:
...
+  before_script: []
+
+eslint-sast:
+  before_script: []
+
+phpcs-security-audit-sast:
+  before_script: []
+
+secrets-sast:
+  before_script: []
```
oliverklee, line 43: Do we need these empty entries?

lukaszuznanski: Yea, because we are running before_script on "default", so we need to clear them in those imported actions

Issue #68, opened 2020-06-25
Rename the *.dist files


oliverklee:
This has two benefits:

- When the file is opened in an IDE, the `.xml` suffix will automatically
  be recognized.
- A `*.dist` file is intended to be copied and modified. This is not what
  we want to communicate here.

PR #112, opened 2020-10-03
[TASK] Rename dist files

lukaszuznanski:
Resolves: #68


oliverklee (changes requested):
There's also `.php_cs.dist ` (a PHP file) to rename.

composer.json
```diff
@@ -68,7 +68,7 @@
     },
     "scripts": {
         "ci:php:lint": "find *.php Classes Configuration Tests -name '*.php' -print0 | xargs -0 -n 1 -P 4 php -l",
-        "ci:php:codestyle": "php-cs-fixer fix --config .php_cs.dist -v --dry-run --using-cache false --diff --diff-format=udiff",
+        "ci:php:codestyle": "php-cs-fixer fix --config .php_cs.xml -v --dry-run --using-cache false --diff --diff-format=udiff",
```
oliverklee, line 71: This file is PHP, not XML. Hence, the name needs to be `.php_cs.php`.

oliverklee (changes requested):

composer.json
lukaszuznanski: Sorry, late night changes :) fix pushed

oliverklee (changes requested):
We'll also need to update the corresponding lines in the `.gitattributes` file. (I'm sorry that I missed that in my earlier reviews.)

oliverklee (approved):

PR #113, opened 2020-10-07
[FEATURE] Add ter release via github actions

lukaszuznanski:


.github/workflows/release.yml
```diff
@@ -0,0 +1,52 @@
...
+          path: ${{ env.TYPO3_EXTENSION_KEY }}
+
+      - uses: shivammathur/setup-php@v2
+        with:
+          php-version: ${{ matrix.php-versions }}
+          extensions: intl, mbstring, xml, soap, zip, curl
+
+      - name: Install TYPO3 TER client
+
+        run: composer global require namelesscoder/typo3-repository-client
```
oliverklee, line 39: As far I know, https://packagist.org/packages/helhum/ter-client builds on `namelesscoder/typo3-repository-client` and is maintained better (at least is has more recent updates). I propose changing to it.

.github/workflows/release.yml
```diff
@@ -0,0 +1,52 @@
...
+          "^2.0" --prefer-dist --no-progress --no-suggest
+
+      - name: Remove .git folder
+        run: rm -rf ./$TYPO3_EXTENSION_KEY/.git
```
oliverklee, line 43: We should remove more - including the tests, development configuration files etc. (Helmut's TER client automatically excludes any dot files.)

Please feel free to steal from one of my extensions and add anything development-only that only tea has:
https://github.com/oliverklee/ext-oelib/blob/master/.travis.yml#L99

.github/workflows/release.yml
```diff
@@ -0,0 +1,52 @@
...
+        with:
+          path: ${{ env.TYPO3_EXTENSION_KEY }}
+
+      - uses: shivammathur/setup-php@v2
+        with:
+          php-version: ${{ matrix.php-versions }}
+          extensions: intl, mbstring, xml, soap, zip, curl
+
+      - name: Install TYPO3 TER client
+
```
oliverklee, line 38: I think not having this blank lines would make it easier to see the blocks.

.github/workflows/release.yml
```diff
@@ -0,0 +1,52 @@
...
+        run: composer global require namelesscoder/typo3-repository-client
+          "^2.0" --prefer-dist --no-progress --no-suggest
+
+      - name: Remove .git folder
+        run: rm -rf ./$TYPO3_EXTENSION_KEY/.git
+
+      - name: List extension folder
+        run: ls -liAsh ./$TYPO3_EXTENSION_KEY
```
oliverklee, line 46: Shouldn't the current directory be the extension directory (i.e., the Git repository root) by default instead?

.github/workflows/release.yml
```diff
@@ -0,0 +1,52 @@
...
+
+      - name: Upload EXT:${{ env.TYPO3_EXTENSION_KEY }}
```
oliverklee, line 48: Is there any way we could read this from the `composer.json` (where the key already should be listed)?

oliverklee (changes requested):

.github/workflows/release.yml
lukaszuznanski: Ok, I'll switch to this package.

PR #74, opened 2020-07-21
[FEATURE] Add frontend linting

lukaszuznanski:
Add eslint with recommended config for typescript
Add styleslint with recommended config

PR #120, opened 2020-10-12
[FEATURE] Add frontend linting

lukaszuznanski:
This PR is the resurrection of #74 (after the switch from master to main).


Resources/Public/JavaScript/Main.js
```diff
@@ -0,0 +1,36 @@
+document.addEventListener("DOMContentLoaded", function () {
+    var t = document.querySelectorAll(".tx-tea table"),
+        i = t.length;
+    while (--i >= 0) makeSortable(t[i]);
+});
+
+function makeSortable(table) {
```
oliverklee, line 7: Should we namespace the functions to avoid name collisions, e.g., like this?
https://github.com/oliverklee/ext-seminars/blob/master/Resources/Public/JavaScript/FrontEnd/FrontEnd.js#L22

Resources/Public/JavaScript/Main.js
```diff
@@ -0,0 +1,36 @@
+document.addEventListener("DOMContentLoaded", function () {
+    var t = document.querySelectorAll(".tx-tea table"),
+        i = t.length;
+    while (--i >= 0) makeSortable(t[i]);
```
oliverklee, line 4: Please let's have the `{` and `}` here. (Shouldn't eslint have complained about this?)

Resources/Public/JavaScript/Main.js
```diff
@@ -0,0 +1,36 @@
...
+        (function (i) {
+            var dir = 1;
+            th[i].addEventListener("click", function () {
+                sortTable(table, i, (dir = 1 - dir));
+            });
+        })(i);
+}
+
+function sortTable(table, col, reverse) {
+    var tb = table.tBodies[0], //
```
oliverklee, line 23: Why the `//` here?

Resources/Private/Templates/Tea/Index.html
```diff
@@ -8,6 +8,12 @@ <h2>
         </h2>
 
         <table class="table">
+            <thead>
+                <tr>
+                    <th>a</th>
+                    <th>b</th>
```
oliverklee, line 14: Please let's have meaningful headings here. I propose the localized labels "UID" and "Title".

.gitattributes
```diff
@@ -7,4 +7,10 @@
 /.php_cs.php export-ignore
 /Tests/ export-ignore
 /codeception.yml export-ignore
+/phpcs.xml.dist export-ignore
 /phpcs.xml export-ignore
+/Resources/Private/.eslintignore export-ignore
+/Resources/Private/.eslintrc export-ignore
```
oliverklee, line 13: This needs to be `/Resources/Private/.eslintrc.json` (with the added `.json`).

Resources/Private/.htaccess
```diff
@@ -1,3 +1,4 @@
+
```
oliverklee, line 1: Why this change?

oliverklee (changes requested):
I have carried over my comments from #74.

Resources/Private/Templates/Tea/Index.html
lukaszuznanski: Looks like I submitted "test" content :)

Resources/Private/.htaccess
lukaszuznanski: miss click :)

oliverklee (changes requested):
The JS linter has something to complain about. :-)

I also have updated the German translation with an additional commit.

.gitattributes
```diff
@@ -7,4 +7,10 @@
 /.php_cs.php export-ignore
 /Tests/ export-ignore
 /codeception.yml export-ignore
+/phpcs.xml.dist export-ignore
```
oliverklee, line 10: We don't have this file anymore.

lukaszuznanski:
> The JS linter has something to complain about. :-)
> 
> I also have updated the German translation with an additional commit.

done :)

oliverklee (approved):
Looks good!

PR #124, opened 2020-10-14
[FEATURE] Add TER release via GitHub actions

lukaszuznanski:
This PR is the resurrection of #113 (after the switch from master to main).

### DIFF
--- a/.github/workflows/release.yml
+++ b/.github/workflows/release.yml
@@ -11,7 +11,6 @@ jobs:
   ter-release:
     name: "TYPO3 TER release"
 
-    # Run only if it's a tag in general (as regex is not possible here)
     if: startsWith(github.ref, 'refs/tags/')
 
     runs-on: ubuntu-latest
@@ -35,15 +34,14 @@ jobs:
           extensions: intl, mbstring, xml, soap, zip, curl
 
       - name: Install TYPO3 TER client
-
-        run: composer global require namelesscoder/typo3-repository-client
-          "^2.0" --prefer-dist --no-progress --no-suggest
+        run: composer global require helhum/ter-client
+          "^0.1" --prefer-dist --no-progress --no-suggest
 
       - name: Remove .git folder
-        run: rm -rf ./$TYPO3_EXTENSION_KEY/.git
+        run: rm -rf CODE _OF_CONDUCT.md Tests codeception.xml phpcs.xml
 
       - name: List extension folder
-        run: ls -liAsh ./$TYPO3_EXTENSION_KEY
+        run: ls -liAsh ./
 
       - name: Upload EXT:${{ env.TYPO3_EXTENSION_KEY }}
           as ${{ github.event.ref }} to TER
--- a/.github/workflows/release.yml
+++ b/.github/workflows/release.yml
@@ -35,16 +35,16 @@ jobs:
 
       - name: Install TYPO3 TER client
         run: composer global require helhum/ter-client
-          "^0.1" --prefer-dist --no-progress --no-suggest
+          "^0.1" --prefer-dist --no-progress
 
-      - name: Remove .git folder
-        run: rm -rf CODE _OF_CONDUCT.md Tests codeception.xml phpcs.xml
+      - name: Remove files
+        run: rm -rf CODE_OF_CONDUCT.md Tests codeception.yml phpcs.xml .github .gitlab .editor .Build .gitattributes .editorconfig .gitignore .php_cs.php Resources/Private/.eslintignore Resources/Private/.eslintrc.json Resources/Private/.prettierrc.js Resources/Private/package.json Resources/Private/stylelint.config.js
 
       - name: List extension folder
         run: ls -liAsh ./
 
       - name: Upload EXT:${{ env.TYPO3_EXTENSION_KEY }}
           as ${{ github.event.ref }} to TER
-        run: php ~/.composer/vendor/bin/upload "./$TYPO3_EXTENSION_KEY"
-          "$TYPO3_USER" "$TYPO3_PASSWORD"
-          "${{ github.event.head_commit.message }}"
+        run: ter-client upload $TYPO3_EXTENSION_KEY "./$TYPO3_EXTENSION_KEY"
+          -u $TYPO3_USER -p $TYPO3_PASSWORD"
+          -m "${{ github.event.head_commit.message }}"
--- a/.github/workflows/release.yml
+++ b/.github/workflows/release.yml
@@ -1,0 +1,52 @@
+---
+name: "TER deployment"
+
+on:
+  push:
+    tags:
+      - 'v[0-9]+.[0-9]+.[0-9]+'
+
+jobs:
+  # The actual TER upload job
+  ter-release:
+    name: "TYPO3 TER release"
+
+    # Run only if it's a tag in general (as regex is not possible here)
+    if: startsWith(github.ref, 'refs/tags/')
+
+    runs-on: ubuntu-latest
+    strategy:
+      matrix:
+        php-versions: [7.4]
+
+    env:
+      TYPO3_EXTENSION_KEY: tea
+      TYPO3_USER: ${{ secrets.TYPO3_USER }}
+      TYPO3_PASSWORD: ${{ secrets.TYPO3_PASSWORD }}
+
+    steps:
+      - uses: actions/checkout@v2
+        with:
+          path: ${{ env.TYPO3_EXTENSION_KEY }}
+
+      - uses: shivammathur/setup-php@v2
+        with:
+          php-version: ${{ matrix.php-versions }}
+          extensions: intl, mbstring, xml, soap, zip, curl
+
+      - name: Install TYPO3 TER client
+
+        run: composer global require namelesscoder/typo3-repository-client
+          "^2.0" --prefer-dist --no-progress --no-suggest
+
+      - name: Remove .git folder
+        run: rm -rf ./$TYPO3_EXTENSION_KEY/.git
+
+      - name: List extension folder
+        run: ls -liAsh ./$TYPO3_EXTENSION_KEY
+
+      - name: Upload EXT:${{ env.TYPO3_EXTENSION_KEY }}
+          as ${{ github.event.ref }} to TER
+        run: php ~/.composer/vendor/bin/upload "./$TYPO3_EXTENSION_KEY"
+          "$TYPO3_USER" "$TYPO3_PASSWORD"
+          "${{ github.event.head_commit.message }}"
--- a/.github/workflows/release.yml
+++ b/.github/workflows/release.yml
@@ -13,10 +13,10 @@ jobs:
 
     if: startsWith(github.ref, 'refs/tags/')
 
-    runs-on: ubuntu-latest
+    runs-on: ubuntu-20.04
     strategy:
       matrix:
-        php-versions: [7.4]
+        php-versions: [ 7.4 ]
 
     env:
       TYPO3_EXTENSION_KEY: tea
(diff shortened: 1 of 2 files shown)


.github/workflows/release.yml
```diff
@@ -0,0 +1,52 @@
...
+          path: ${{ env.TYPO3_EXTENSION_KEY }}
+
+      - uses: shivammathur/setup-php@v2
+        with:
+          php-version: ${{ matrix.php-versions }}
+          extensions: intl, mbstring, xml, soap, zip, curl
+
+      - name: Install TYPO3 TER client
+
+        run: composer global require namelesscoder/typo3-repository-client
```
oliverklee, line 39: As discussed in https://github.com/TYPO3-Documentation/tea/pull/113#discussion_r500963090, we should switch to https://packagist.org/packages/helhum/ter-client.

.github/workflows/release.yml
```diff
@@ -0,0 +1,52 @@
...
+          "^2.0" --prefer-dist --no-progress --no-suggest
+
+      - name: Remove .git folder
+        run: rm -rf ./$TYPO3_EXTENSION_KEY/.git
```
oliverklee, line 43: We should remove more - including the tests, development configuration files etc. (Helmut's TER client automatically excludes any dot files: https://github.com/helhum/ter-client/blob/master/src/ExtensionUploadPacker.php#L235)

Please feel free to steal from one of my extensions and add anything development-only that only tea has:
https://github.com/oliverklee/ext-oelib/blob/master/.travis.yml#L99

.github/workflows/release.yml
```diff
@@ -0,0 +1,52 @@
...
+        with:
+          path: ${{ env.TYPO3_EXTENSION_KEY }}
+
+      - uses: shivammathur/setup-php@v2
+        with:
+          php-version: ${{ matrix.php-versions }}
+          extensions: intl, mbstring, xml, soap, zip, curl
+
+      - name: Install TYPO3 TER client
+
```
oliverklee, line 38: I think not having this blank lines would make it easier to see the blocks.

.github/workflows/release.yml
```diff
@@ -0,0 +1,52 @@
...
+        run: composer global require namelesscoder/typo3-repository-client
+          "^2.0" --prefer-dist --no-progress --no-suggest
+
+      - name: Remove .git folder
+        run: rm -rf ./$TYPO3_EXTENSION_KEY/.git
+
+      - name: List extension folder
+        run: ls -liAsh ./$TYPO3_EXTENSION_KEY
```
oliverklee, line 46: Shouldn't the current directory be the extension directory (i.e., the Git repository root) by default instead?

.github/workflows/release.yml
```diff
@@ -0,0 +1,52 @@
...
+
+      - name: Upload EXT:${{ env.TYPO3_EXTENSION_KEY }}
```
oliverklee, line 48: Is there any way we could read this from the composer.json (where the key already should be listed)?

oliverklee (changes requested):
I've carried over my comments from #113.

.github/workflows/release.yml
```diff
@@ -0,0 +1,50 @@
...
+          "^0.1" --prefer-dist --no-progress
+
+      - name: Remove files
+        run: rm -rf CODE_OF_CONDUCT.md Tests codeception.yml phpcs.xml .github .gitlab .editor .Build .gitattributes .editorconfig .gitignore .php_cs.php Resources/Private/.eslintignore Resources/Private/.eslintrc.json Resources/Private/.prettierrc.js Resources/Private/package.json Resources/Private/stylelint.config.js
+
+      - name: List extension folder
+        run: ls -liAsh ./
+
+      - name: Upload EXT:${{ env.TYPO3_EXTENSION_KEY }}
+          as ${{ github.event.ref }} to TER
```
oliverklee, line 47: I think we need to use the `>` operator for multi-line commands (or put it all in once line), e.g., like this:

https://github.com/oliverklee/ext-phpunit/blob/main/.github/workflows/ci.yml#L120

(I'm not completely sure that this is necessary, though).

.github/workflows/release.yml
```diff
@@ -0,0 +1,50 @@
...
+        run: ter-client upload $TYPO3_EXTENSION_KEY "./$TYPO3_EXTENSION_KEY"
+          -u $TYPO3_USER -p $TYPO3_PASSWORD"
+          -m "${{ github.event.head_commit.message }}"
```
oliverklee, line 50: Currently, we're using the commit message of the last commit that was tagged. Instead, we should use the commit message of the (annotated) tag.

.github/workflows/release.yml
```diff
@@ -0,0 +1,50 @@
...
+      - name: Remove files
+        run: rm -rf CODE_OF_CONDUCT.md Tests codeception.yml phpcs.xml .github .gitlab .editor .Build .gitattributes .editorconfig .gitignore .php_cs.php Resources/Private/.eslintignore Resources/Private/.eslintrc.json Resources/Private/.prettierrc.js Resources/Private/package.json Resources/Private/stylelint.config.js
+
+      - name: List extension folder
+        run: ls -liAsh ./
+
+      - name: Upload EXT:${{ env.TYPO3_EXTENSION_KEY }}
+          as ${{ github.event.ref }} to TER
+        run: ter-client upload $TYPO3_EXTENSION_KEY "./$TYPO3_EXTENSION_KEY"
+          -u $TYPO3_USER -p $TYPO3_PASSWORD"
```
oliverklee, line 49: There's an uneven number of double quotes here.

oliverklee (changes requested):

.github/workflows/release.yml
lukaszuznanski: Yea it's still a draft, will work on it in the following days

.github/workflows/release.yml
```diff
@@ -0,0 +1,50 @@
...
+          path: ${{ env.TYPO3_EXTENSION_KEY }}
+
+      - uses: shivammathur/setup-php@v2
+        with:
+          php-version: ${{ matrix.php-versions }}
+          extensions: intl, mbstring, xml, soap, zip, curl
+
+      - name: Install TYPO3 TER client
+        run: composer global require helhum/ter-client
+          "^0.1" --prefer-dist --no-progress
```
oliverklee, line 38: As version 0.1.1 has been released, we can use `^0.1.1` now.

.github/workflows/release.yml
lukaszuznanski: Unfortunately, I can't find any reasonable way to read it from release description, do you have any idea how to get this data?

.github/workflows/release.yml
```diff
@@ -0,0 +1,65 @@
+---
+name: "TER deployment"
+
+on:
+  push:
+    tags:
+      - 'v[0-9]+.[0-9]+.[0-9]+'
```
oliverklee, line 7: By the way, would it make sense to use the [release event](https://docs.github.com/en/free-pro-team@latest/actions/reference/events-that-trigger-workflows#release) instead?

.github/workflows/release.yml
oliverklee: Maybe we can take the [release event payload](https://docs.github.com/en/free-pro-team@latest/rest/reference/repos#get-a-release) and use the first line of the release description?

oliverklee:
With tailor: https://github.com/o-ba/tailor_ext/blob/main/.github/workflows/publish.yml

.github/workflows/release.yml
```diff
@@ -0,0 +1,81 @@
...
+            echo ::set-output name=comment::Released version ${{ steps.get-version.outputs.version }} of ${{ steps.split.outputs._1 }}
+          else
+            echo ::set-output name=comment::$comment
+          fi
+
+      - name: Setup PHP
+        uses: shivammathur/setup-php@v2
+        with:
+          php-version: 7.4
+          extensions: intl, mbstring, json, zip, curl
```
oliverklee, line 75: Please let's set the Composer version (2) explicitly (like [in the CI workflow](https://github.com/TYPO3-Documentation/tea/blob/main/.github/workflows/ci.yml#L42)).

.github/workflows/release.yml
```diff
@@ -0,0 +1,81 @@
...
+  ter-release:
+    needs: set-matrix
+    name: "TYPO3 TER release"
+
+    if: startsWith(github.ref, 'refs/tags/')
+
+    runs-on: ubuntu-20.04
+    strategy:
+      matrix:
+        php-versions: [ 7.4 ]
```
oliverklee, line 29: We're hardcoding the the PHP version in line 74. So there is no need to have the PHP version in a matrix.

.github/workflows/release.yml
```diff
@@ -0,0 +1,81 @@
+---
+name: "TER deployment"
```
oliverklee, line 2: Please let's synchronize the name of the workflow and the file name. I propose to use "publish" for both so we are close to https://github.com/o-ba/tailor_ext/blob/main/.github/workflows/publish.yml.

.github/workflows/release.yml
```diff
@@ -0,0 +1,81 @@
...
+jobs:
+  set-matrix:
+    runs-on: ubuntu-20.04
+    outputs:
+      matrix: ${{ steps.set-matrix.outputs.matrix }}
+    steps:
+      - id: set-matrix
+        run: |
+          JSON=$(cat ./composer.json)
+          echo "::set-output name=${JSON//'%'/'%25'}"
```
oliverklee, line 17: What does this job do? (I don't understand it yet …)

.github/workflows/release.yml
```diff
@@ -0,0 +1,81 @@
...
+
+on:
+  release:
+    types: [published]
+
+jobs:
+  set-matrix:
+    runs-on: ubuntu-20.04
+    outputs:
+      matrix: ${{ steps.set-matrix.outputs.matrix }}
```
oliverklee, line 12: Should the output really be a matrix, or should/can it be an env var instead?

.github/workflows/release.yml
```diff
@@ -0,0 +1,81 @@
...
+        id: split
+        with:
+          msg: ${{ matrix.composer-data.name}}
+          separator: '/'
+
+      - name: Remove files
+        run: >
+          rm -rf CODE_OF_CONDUCT.md Tests codeception.yml phpcs.xml .github .gitlab .editor .Build .gitattributes
+          .editorconfig .gitignore .php_cs.php Resources/Private/.eslintignore Resources/Private/.eslintrc.json
+          Resources/Private/.prettierrc.js Resources/Private/package.json Resources/Private/stylelint.config.js"
```
oliverklee, line 59: What's the trailing double quote doing here?

.github/workflows/release.yml
```diff
@@ -0,0 +1,81 @@
...
+
+      - uses: jungwinter/split@v1
+        id: split
+        with:
+          msg: ${{ matrix.composer-data.name}}
+          separator: '/'
+
+      - name: Remove files
+        run: >
+          rm -rf CODE_OF_CONDUCT.md Tests codeception.yml phpcs.xml .github .gitlab .editor .Build .gitattributes
```
oliverklee, line 57: With the [exclude list included in the latest tailor version](https://github.com/TYPO3/tailor/blob/main/conf/ExcludeFromPackaging.php), we only need to remove the following file (due to a [typo in the exclude list](https://github.com/TYPO3/tailor/pull/19)):
```
.php_cs.php
```

.github/workflows/release.yml
```diff
@@ -0,0 +1,81 @@
...
+    steps:
+      - id: set-matrix
+        run: |
+          JSON=$(cat ./composer.json)
+          echo "::set-output name=${JSON//'%'/'%25'}"
+
+  # The actual TER upload job
+  ter-release:
+    needs: set-matrix
+    name: "TYPO3 TER release"
```
oliverklee, line 22: Please let's have the name line directly below the key (above the `needs` line) so both are together (as they say the same).

.github/workflows/release.yml
```diff
@@ -0,0 +1,81 @@
+---
+name: "TER deployment"
+
+on:
+  release:
+    types: [published]
+
+jobs:
+  set-matrix:
```
oliverklee, line 9: Please let's give this job a name, too.

oliverklee (changes requested):

.github/workflows/release.yml
lukaszuznanski: This job get ext version from composer.json. In order to process json file we need to set it to matrix.

.github/workflows/release.yml
lukaszuznanski: As I commented above

lukaszuznanski:
Solved by: #189

Issue #116, opened 2020-10-10
Unify the sniffing and fixing Composer commands


oliverklee:
We should have only one command to do both sniffs and both fixers:
- `ci:php:sniff`
- `php:fix`

PR #147, opened 2020-10-20
[TASK] Add one command to do sniffers and fixers

lukaszuznanski:
I'm not sure if this is what you meant in issue, let me know :).

Resolves: #116

### DIFF
--- a/composer.json
+++ b/composer.json
@@ -81,6 +81,10 @@
         "ci:dynamic": [
             "@ci:tests"
         ],
+        "ci:php": [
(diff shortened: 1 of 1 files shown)


oliverklee:
@lukaszuznanski I was thinking of something a bit different. I'll update the PR in a second. (I think that's faster than me trying to explain this in words. ;-) )

oliverklee:
My bad: I misread the PR, and I take everything back. This is exactly what I meant.

oliverklee (approved):

PR #149, opened 2020-10-20
[FEATURE] Add phpcpd

lukaszuznanski:
Resolves: #129


composer.json
```diff
@@ -69,6 +70,7 @@
         "ci:php:lint": "find *.php Classes Configuration Tests -name '*.php' -print0 | xargs -0 -n 1 -P 4 php -l",
         "ci:php:codestyle": "php-cs-fixer fix --config .php_cs.php -v --dry-run --using-cache false --diff --diff-format=udiff",
         "ci:php:sniff": ".Build/vendor/bin/phpcs Classes Configuration Tests",
+        "ci:php:copypaste": ".Build/vendor/bin/phpcpd .",
```
oliverklee, line 73: Won't this also check `vendor/`?

And I'm also not sure we should the TCA files. Maybe only check `Classes/` and `Tests/`? What do you think?

lukaszuznanski: Yea probably this will also test .Build/vendor directory. I will check how this checker works with TCA code (if it's seen as copy paste) and then I will let you know.

oliverklee (changes requested):
Looks good to me. Could you please resolve the merge conflict and also add this new feature to the feature list?

oliverklee (approved):

PR #156, opened 2020-10-31
Migrate secret detection in gitlab ci

lukaszuznanski:
secrets-sast was removed from SAST vendored template
https://gitlab.com/gitlab-org/gitlab/-/issues/234011

we need to migrate to separate template, it's breaking gitlab ci template ATM, as it's moved out and it won't work anymore

### DIFF
--- a/.gitlab/pipeline/.gitlab-ci.yml
+++ b/.gitlab/pipeline/.gitlab-ci.yml
@@ -128,6 +128,7 @@ func-php7.2-v10:
     - build-composer-dependencies
     - php-lint-php7.2
   script:
+    - composer require typo3/minimal:"^10.4"
     - composer ci:tests:functional
 
 func-php7.3-v10:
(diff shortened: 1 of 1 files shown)


oliverklee:
This branch currently is failing on GitLab CI: https://gitlab.typo3.org/qa/example-extension/-/pipelines/9827

lukaszuznanski:
Yea but its failing due to composer dependencies. I will rerun this pipeline after fixing 
 php7.2 dependencies issue. 
Some of pull requests created regression, but we were not able to check which as gitlab ci file is not working fine right now (as this job was moved out and because of this, now syntax is incorrect).

lukaszuznanski:
I added composer require t3 v10 to tests (as it is on github actions already), but I can't move it to separate branch, as we can't test it this way. 

Here is pipeline https://gitlab.typo3.org/qa/example-extension/-/pipelines/9942

oliverklee (approved):

PR #160, opened 2020-11-18
[FEATURE] Also run the CI build once a week

oliverklee:
This will allow us to catch problems with dependencies or
the build environment, i.e., problems that are not related
to changes in our code base.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -7,6 +7,8 @@ on:
   pull_request:
     branches:
       - main
+  schedule:
+    - cron: '15 3 * * 1'
 jobs:
   php-lint:
     name: "PHP linter"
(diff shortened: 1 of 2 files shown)


lukaszuznanski:
Looks fine, do we also need to setup gitlab CI schedule? This requires setup in gitlab https://docs.gitlab.com/ee/ci/pipelines/schedules.html

lukaszuznanski (commented):
Looks fine, do we also need to setup gitlab CI schedule? This requires setup in gitlab https://docs.gitlab.com/ee/ci/pipelines/schedules.html

oliverklee:
> Looks fine, do we also need to setup gitlab CI schedule?

Possibly. I'd like to do that in a separate PR, though. Would you be willing to do that?

lukaszuznanski:
Ok, it's only setup in gitlab repository, so I will create issue and then add it in gitlab. Let's merge this one.

lukaszuznanski (approved):

PR #162, opened 2020-11-21
[BUGFIX] Explicitly set the Ubuntu version for GitHub Actions

oliverklee:
This fixes a warning when running the CI build.

As the functional tests currently do not run with the Ubuntu 20.04
runner, we need to keep using 18.04.
    
Also explicitly require the `mysqli` PHP extension.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -166,7 +166,7 @@ jobs:
           - ^10.4
   functional-tests:
     name: "Functional tests"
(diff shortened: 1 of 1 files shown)


oliverklee:
This breaks the functional tests. I will need to investigate this.

oliverklee:
I have moved upgrading to the Ubuntu 20.04 runner to #173.

lukaszuznanski (approved):

PR #176, opened 2020-11-29
[CLEANUP] Adopt .editorconfig settings from the TYPO3 core

oliverklee:
Also drop redundancies and sort the entries.

### DIFF
--- a/.editorconfig
+++ b/.editorconfig
@@ -5,66 +5,45 @@ root = true
 [*]
 charset = utf-8
 end_of_line = lf
+indent_style = space
+indent_size = 4
 insert_final_newline = true
 trim_trailing_whitespace = true
 
-# HTML files
-[*.html]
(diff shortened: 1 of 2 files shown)


.editorconfig
```diff
@@ -5,66 +5,45 @@ root = true
...
-indent_size = 4
+# SQL files
+[*.sql]
+indent_style = tab
+indent_size = 2
 
-# YML files
-[{*.yml,*.yaml}]
-indent_style = space
+# TypoScript file
```
lukaszuznanski, line 35: I think we should keep 'files' like in every line, instead of 'file' :)

oliverklee: Good catch! Fixed. (I'll still need to fix the build failures before this PR is ready for review, though.)

lukaszuznanski (approved):

PR #190, opened 2021-02-03
Use PHPCPD as phar instead of composer package

lukaszuznanski:


.github/workflows/ci.yml
```diff
@@ -60,9 +60,23 @@ jobs:
...
+        uses: actions/checkout@v2
+      - name: "Install PHP"
+        uses: shivammathur/setup-php@v2
+        with:
+          php-version: 7.4
+          coverage: none
+      - name: "Download tool"
+        run: "wget https://phar.phpunit.de/phpcpd.phar"
+      - name: "Run command"
+        run: "php phpcpd.phar --fuzzy Classes Configuration Tests"
```
oliverklee, line 79: Instead of downloading the PHAR on every CI run, we should put the PHAR in `tools/` (via PHIVE) and exclude that directory via `.gitattributes` from releases.

oliverklee (changes requested):

.github/workflows/ci.yml
oliverklee: Like here: https://github.com/MyIntervals/emogrifier

.gitattributes
```diff
@@ -11,5 +11,6 @@
 /Resources/Private/package.json export-ignore
 /Resources/Private/stylelint.config.js export-ignore
 /Tests/ export-ignore
+/tools/ export-ignore
```
oliverklee, line 14: Could you please re-sort the file (using PhpStorm Edit > sort lines)?

.github/workflows/ci.yml
```diff
@@ -60,9 +60,21 @@ jobs:
           - "json:lint"
           - "php:sniff"
           - "php:codestyle"
-          - "php:copypaste"
```
oliverklee, line 63: Please let's modify the corresponding Composer command instead of adding a new job.

oliverklee (changes requested):

oliverklee (approved):

PR #206, opened 2021-02-24
[BUGFIX] Downgrade the XLIFF files to version 1.0

### LINKED ISSUE #198: How to deal with mismatching XLF versions in tea, documentation and TYPO3 core
I've created this PR https://github.com/TYPO3-Documentation/TYPO3CMS-Reference-CoreApi/pull/1301 where I wanted to update the xliff version number in the example code in the documentation to 1.2, in order to reflect the best practice which is stated in this example extension here, since both belong to the documentation team: https://github.com/TYPO3-Documentation/tea/blob/main/Resources/Private/Language/locallang.xlf.

My PR got declined because TYPO3 Core doesn't use its own best practices, so the reasoning was not to update the example code. I understand that a mismatch between best practice and what the core does is not desirable. Ideally, those two whings should be in line. Still I would like to know what the rule is for documentation - do we only refer to best practices if they are obeyed in TYPO3 core? Or do we allow the documentation to differ between what's currently happening in TYPO3 and how things _should_ be by pointing to this extension here?

Either way, discussing that in the documentation channel, the question was raised whether XLF 1.2 is actually is considered best practice or not, and it was suggested to change the best practice here back to 1.0 or 1.1. 

Ideally, if we don't want to change the recommendation here, the TYPO3 language files should be updated to 1.2 and documentation also. In that case I'd create a ticket for TYPO3 core.

oliverklee:
This now matches what the TYPO3 Core uses.

We still use the XLIFF 1.2 schema for validation as version 1.2
is the first fully ratified version.

Closes #198

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -95,8 +95,7 @@ jobs:
       - name: "Download xliff schema"
         run: "wget https://docs.oasis-open.org/xliff/v1.2/os/xliff-core-1.2-strict.xsd"
       - name: "Run lint"
-        run: "xmllint --schema ./xliff-core-1.2-strict.xsd
-              --noout $(find Resources -name '*.xlf')"
+        run: "xmllint --schema ./xliff-core-1.2-strict.xsd --noout $(find Resources -name '*.xlf')"
   unit-tests:
     name: "Unit tests"
     runs-on: ubuntu-20.04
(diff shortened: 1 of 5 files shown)


lukaszuznanski:
This requires also change to xliff schema

oliverklee:
Thanks, on it.

oliverklee:
It looks like there is no 1.0 schema available:
https://docs.oasis-open.org/xliff/

I'm going to try whether the file validates with the 1.1 XSD (which seems to be available).

oliverklee:
Okay, this is kind of hacky: We define the files as being version 1.0, but use the 1.1 namespace for validation.

lukaszuznanski (approved):

PR #207, opened 2021-02-24
[TASK] Migrate yaml-lint to PHIVE

### LINKED ISSUE #201: Migrate yaml-lint to phive
j13k/yaml-lint package is missing GPG signature and it's stopping us from using it via PHAR. I created issue https://github.com/j13k/yaml-lint/issues/29

oliverklee:
Fixes #201


oliverklee:
yaml-lint currently now also checks `Resources/Private/node_modules/`, and there's an invalid file there. We'll need to exclude that directory first:

lukaszuznanski (approved):

PR #211, opened 2021-03-10
Add test execution via docker

lukaszuznanski:


lukaszuznanski:
Use ddev instead of docker setup.

PR #220, opened 2021-03-20
[TASK] Migrate TypoScript linting to a GitHub action

lukaszuznanski:
I prepared github action for typoscript linting, migrate tea extension to use it as example
https://github.com/TYPO3-Continuous-Integration/TYPO3-CI-Typoscript-Lint

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -36,7 +36,7 @@ jobs:
       - name: "Checkout"
         uses: actions/checkout@v2
       - name: "Run TypoScript lint"
-        uses: typo3-continous-integration/typo3-ci-typoscript-lint@v1
+        uses: TYPO3-Continuous-Integration/TYPO3-CI-Typoscript-Lint@v1
     strategy:
       fail-fast: false
   code-quality:
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -29,6 +29,16 @@ jobs:
           - 7.2
           - 7.3
           - 7.4
+  typoscript-lint:
+    name: "TypoScript linter"
+    runs-on: ubuntu-20.04
+    steps:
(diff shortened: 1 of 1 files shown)


oliverklee (changes requested):
We also need to remove `tools/typo3-typoscript-lint` and remove it from the PHIVE configuration file.

.github/workflows/ci.yml
```diff
@@ -29,6 +29,16 @@ jobs:
...
+  typoscript-lint:
+    name: "TypoScript linter"
+    runs-on: ubuntu-20.04
+    steps:
+      - name: "Checkout"
+        uses: actions/checkout@v2
+      - name: "Run TypoScript lint"
+        uses: TYPO3-Continuous-Integration/TYPO3-CI-Typoscript-Lint@v1
+    strategy:
+      fail-fast: false
```
oliverklee, line 41: This is not needed in this case it this setting does not make any difference for jobs without a matrix.

lukaszuznanski:
> We also need to remove `tools/typo3-typoscript-lint` and remove it from the PHIVE configuration file.

Maybe we should keep it in phive and in composer.json in order to run lints locally?

oliverklee:
> Maybe we should keep it in phive and in `composer.json` in order to run lints locally?

I hadn't thought of this … is there any way we can avoid these redundancies? Maybe execute the GitHub actions locally?

lukaszuznanski:
> > Maybe we should keep it in phive and in `composer.json` in order to run lints locally?
> 
> I hadn't thought of this … is there any way we can avoid these redundancies? Maybe execute the GitHub actions locally?

Right now I found only "nektos/act" tool to running github actions locally, but anyway, removal of phar files and composer.json scripts will break gitlab-ci pipeline.

oliverklee (approved):

PR #223, opened 2021-03-24
[TASK] Migrate to github actions codestyle check

lukaszuznanski:
https://github.com/TYPO3-Continuous-Integration/TYPO3-CI-Codestyle

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -94,7 +94,6 @@ jobs:
           - "json:lint"
           - "php:cs-fixer"
           - "php:sniff"
-          - "php:copypaste"
           - "php:stan"
         php-version:
           - 7.4
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -45,6 +45,14 @@ jobs:
         uses: actions/checkout@v2
       - name: "Run YAML lint"
         uses: TYPO3-Continuous-Integration/TYPO3-CI-YAML-Lint@v1
+  codestyle-check:
+    name: "Codestyle-check"
+    runs-on: ubuntu-20.04
+    steps:
+      - name: "Checkout"
+        uses: actions/checkout@v2
+      - name: "Run Copypaste Detector"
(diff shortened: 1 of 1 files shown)


.github/workflows/ci.yml
```diff
@@ -37,6 +37,14 @@ jobs:
         uses: actions/checkout@v2
       - name: "Run TypoScript lint"
         uses: TYPO3-Continuous-Integration/TYPO3-CI-Typoscript-Lint@v1
+  codestyle-check:
+    name: "Codestyle check"
```
oliverklee, line 41: Please let's rename this job (and the new action and the Composer action) to "PHP CS Fixer" (`ci:php:cs-fixer`) to make it more clear what tool this is about.

.github/workflows/ci.yml
```diff
@@ -37,6 +37,14 @@ jobs:
...
+    runs-on: ubuntu-20.04
+    steps:
+      - name: "Checkout"
+        uses: actions/checkout@v2
+      - name: "Run Copypaste Detector"
```
oliverklee, line 46: This looks like a copy'n'paste error. ;-)

.github/workflows/ci.yml
```diff
@@ -37,6 +37,14 @@ jobs:
         uses: actions/checkout@v2
       - name: "Run TypoScript lint"
         uses: TYPO3-Continuous-Integration/TYPO3-CI-Typoscript-Lint@v1
+  codestyle-check:
+    name: "Codestyle check"
+    runs-on: ubuntu-20.04
```
oliverklee, line 42: This job should depend on the PHP linter. (If the PHP is invalid, it does not make sense to check for duplication.)

oliverklee (changes requested):
This also needs rebasing.

.github/workflows/ci.yml
oliverklee: I've done one part in #227.

.github/workflows/ci.yml
lukaszuznanski: that's true :) Thanks!

oliverklee (approved):

PR #236, opened 2021-04-06
[TASK] Remove php occurencies in lint scripts

lukaszuznanski:
Is there any case in which using php-cli is required here? I can execute them locally without any problems. We are not setting php in php-cs-fixer fix anyway. Maybe we could get rid of those?

### DIFF
--- a/composer.json
+++ b/composer.json
@@ -80,7 +80,7 @@
 		"ci:dynamic": [
 			"@ci:tests"
 		],
-		"ci:json:lint": "find . ! -path '*.Build/*' ! -path '*Resources/Private/node_modules/*' -name '*.json' | xargs @php .Build/vendor/bin/jsonlint -q",
+		"ci:json:lint": "find . ! -path '*.Build/*' ! -path '*Resources/Private/node_modules/*' -name '*.json' | xargs php .Build/vendor/bin/jsonlint -q",
 		"ci:php": [
 			"@ci:php:copypaste",
 			"@ci:php:cs-fixer",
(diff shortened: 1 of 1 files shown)


oliverklee (changes requested):
We can either completely remove the `php` (if we don't plan to support development on Windows), or should change this to `@php` (so the PHP process of Composer is used) so we support development on Windows as well. I'm all for the latter version.

composer.json
```diff
@@ -75,23 +75,23 @@
 		"ci": [
 			"@ci:static"
 		],
-		"ci:composer:normalize": "php ./tools/composer-normalize --dry-run",
+		"ci:composer:normalize": "@php ./tools/composer-normalize --dry-run",
 		"ci:dynamic": [
 			"@ci:tests"
 		],
-		"ci:json:lint": "find . ! -path '*.Build/*' ! -path '*Resources/Private/node_modules/*' -name '*.json' | xargs .Build/vendor/bin/jsonlint -q",
+		"ci:json:lint": "find . ! -path '*.Build/*' ! -path '*Resources/Private/node_modules/*' -name '*.json' | xargs @php .Build/vendor/bin/jsonlint -q",
```
oliverklee, line 82: For the linter, we cannot use `@php` as this does not work in pipes. So this needs to stay `php`.

composer.json
```diff
@@ -109,14 +109,14 @@
 		],
 		"ci:tests:functional": "find 'Tests/Functional' -wholename '*Test.php' | parallel --gnu 'echo; echo \"Running functional test suite {}\"; .Build/vendor/bin/phpunit -c .Build/vendor/nimut/testing-framework/res/Configuration/FunctionalTests.xml {}';",
 		"ci:tests:unit": ".Build/vendor/bin/phpunit -c .Build/vendor/nimut/testing-framework/res/Configuration/UnitTests.xml Tests/Unit",
-		"ci:ts:lint": "php ./tools/typo3-typoscript-lint -c Configuration/TsLint.yml --ansi -n --fail-on-warnings -vvv Configuration/TypoScript",
-		"ci:yaml:lint": "find . ! -path '*.Build/*' ! -path '*Resources/Private/node_modules/*' -name '*.yml' | xargs php ./tools/yaml-lint",
+		"ci:ts:lint": "@php ./tools/typo3-typoscript-lint -c Configuration/TsLint.yml --ansi -n --fail-on-warnings -vvv Configuration/TypoScript",
+		"ci:yaml:lint": "find . ! -path '*.Build/*' ! -path '*Resources/Private/node_modules/*' -name '*.yml' | xargs @php ./tools/yaml-lint",
```
oliverklee, line 113: Ditto.

oliverklee (changes requested):

oliverklee (approved):

PR #239, opened 2021-04-07
[TASK] Migrate the PHPStan check to GitHub actions

lukaszuznanski:


oliverklee (changes requested):
This does not take the project `phpstan.neon` into account. However, we need a way to override the GitHub Action's default settings.

lukaszuznanski:
> This does not take the project `phpstan.neon` into account. However, we need a way to override the GitHub Action's default settings.

I added config file to action

oliverklee (approved):

Issue #246, opened 2021-05-02
fix:php:sniff uses the wrong command


oliverklee:
This command sniffs, but it should fix instead.

PR #247, opened 2021-05-03
[BUGFIX] Fix phpcs fix script

lukaszuznanski:
Resolves: #246


composer.json
```diff
@@ -116,7 +116,7 @@
 			"@fix:php:sniff"
 		],
 		"fix:php:cs": "./tools/php-cs-fixer fix --config .php_cs.php",
-		"fix:php:sniff": ".Build/vendor/bin/phpcs Classes Configuration Tests",
+		"fix:php:sniff": ".Build/vendor/bin/phpcbf Classes Configuration Tests",
```
oliverklee, line 119: Please let's add `phpcbf` to the PHIVE-installed tools and use that.

oliverklee (changes requested):

composer.json
oliverklee: Also, we currently do not have PHP_CodeSniffer installed as a Composer dependency.

lukaszuznanski: alright, that's true :)

composer.json
```diff
@@ -117,7 +117,7 @@
...
```
oliverklee, line 120: We'll need to refer to the PHAR here, not the Composer-installed executable.

oliverklee (changes requested):

oliverklee (approved):

Issue #217, opened 2021-03-16
Raise the PHPStan level and fix any issues


oliverklee:

oliverklee:
Fixed with #263. Closing.

PR #263, opened 2021-07-14
[TASK] Raise PHP Stan level to 8

lukaszuznanski:
Resolves: #217

### DIFF
--- a/Classes/Domain/Model/Product/Tea.php
+++ b/Classes/Domain/Model/Product/Tea.php
@@ -54,7 +54,8 @@ class Tea extends AbstractEntity
 
     public function getImage(): ?FileReference
     {
-        if ($this->image instanceof LazyLoadingProxy) { // @phpstan-ignore-line
+        // @phpstan-ignore-next-line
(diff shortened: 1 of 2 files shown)


Classes/Domain/Model/Product/Tea.php
```diff
@@ -54,7 +54,7 @@ public function setDescription(string $description): void
 
     public function getImage(): ?FileReference
     {
-        if ($this->image instanceof LazyLoadingProxy) {
+        if ($this->image instanceof LazyLoadingProxy) { // @phpstan-ignore-line
```
oliverklee: The code sniffer needs this to be on the line above.

But apart from that, I'd prefer to have the issues in a baseline file (like [in this project](https://github.com/sabberworm/PHP-CSS-Parser)) so we have a list of things we probably need to fix.

oliverklee (changes requested):

flicstar:
@oliverklee PING. It is our Sweeping Week when we chase up open PRs. Looks like this one is waiting on a review from you. Just wanted to make sure it's in your queue. Thanks!

PR #271, opened 2021-07-30
[BUGFIX] Fix default template names for GitLab CI

jpmschuler:
GitLab offers default jobs for LicenseManagement, SecretDetection and SAST which are used here. However since some versions the filename of these includes was adapted and harmonised. Running the current files creates no pipelines due to a YAML error, as these includes are not found.


oliverklee (approved):

oliverklee:
Thanks for the fix, @jpmschuler! :heart:

PR #272, opened 2021-08-10
[TASK] Update phive tools and cleanup

lukaszuznanski:


oliverklee (changes requested):
We'll also need the updates to the files in `tools/` for this PR.

oliverklee (approved):
Oh, it's only the cleanup, no updates. I see.

PR #273, opened 2021-08-10
[TASK] Split GitLab jobs

lukaszuznanski:
https://gitlab.typo3.org/qa/example-extension/-/pipelines/15250


lukaszuznanski:
https://gitlab.typo3.org/qa/example-extension/-/pipelines/15250

oliverklee (commented):
Concerning the file structure: I don't understand why the directory is called `templates`. Are those files intended to be copied and changed instead of being used directly?

lukaszuznanski:
> Concerning the file structure: I don't understand why the directory is called `templates`. Are those files intended to be copied and changed instead of being used directly?

Those files should be used directly. I changed directory name.

oliverklee (approved):
LGTM.

PR #293, opened 2021-09-07
[TASK] Start to work on TYPO3 v11 support

lukaszuznanski:
Resolves: #286


oliverklee:
We'll need to add `"ignore-as-root": false,` to the `typo3/cms` section in the `composer.json` so the root package will be recognized as an extension again in TYPO3 >= 11.4: https://github.com/oliverklee/ext-phpunit/pull/266

oliverklee (approved):

Issue #281, opened 2021-08-22
Extend the PHPStan configuration


oliverklee:
move the list of checked folders from the Composer command to the PHPStan configuration

PR #296, opened 2021-09-07
[TASK] Extend phpstan configuraton

lukaszuznanski:
Resolves: #281


oliverklee (changes requested):
The build now is red. Maybe we need to update the baseline.

lukaszuznanski:
> The build now is red. Maybe we need to update the baseline.

Yes, I forgot to set it as draft. I will create separate pull request with fixes.

Issue #291, opened 2021-08-26
Drop the LazyLoading proxy special handling from Tea::getImage


oliverklee:
We need to make sure this is covered with a functional test first.

oliverklee:
It looks like we cannot do this:
```
TypeError : Return value of TTN\Tea\Domain\Model\Product\Tea::getImage() must be an instance of TYPO3\CMS\Extbase\Domain\Model\FileReference or null, instance of TYPO3\CMS\Extbase\Persistence\Generic\LazyLoadingProxy returned
 /home/klee/src/typo3/ext/tea/Classes/Domain/Model/Product/Tea.php:59
 /home/klee/src/typo3/ext/tea/Tests/Functional/Domain/Repository/Product/TeaRepositoryTest.php:108
```

Closing.

PR #300, opened 2021-09-10
[CLEANUP] Make lazy-loading proxies visible via type annotations

oliverklee:
Relates to #291


oliverklee:
Marking as draft again. I think we should have first #303.

PR #319, opened 2021-10-06
[FEATURE] Also run the tests with TYPO3 11LTS in GitHub Actions

oliverklee:


oliverklee:
We probably either need to wait for https://github.com/helhum/typo3-composer-setup to be 11-compatible, or try to do without it.

lukaszuznanski (approved):

Issue #320, opened 2021-10-06
Use schnitzler/phpstan-typo3-extbase 


oliverklee:
https://packagist.org/packages/schnitzler/phpstan-typo3-extbase

oliverklee:
Closing, as this has turned out to be not feasible for our setup (see https://github.com/TYPO3-Documentation/tea/pull/332#pullrequestreview-796288731 for details).

PR #332, opened 2021-11-03
[TASK] Add phpstan typo3 extbase plugin

lukaszuznanski:
Resolves: #320


oliverklee (changes requested):
This package seeds to only work with TYPO3 10 (no 9, no 11). So maybe using it won't work for us (as we want to support all current TYPO3 LTS versions).

lukaszuznanski:
Ok, closing for now.

PR #338, opened 2021-11-12
[TASK] allow newlines for ternary operator

### LINKED ISSUE #337: Allow newlines in ternary
I've just stumbled over the same as in https://github.com/squizlabs/PHP_CodeSniffer/issues/3377#issuecomment-860028311 and have seen that @oliverklee had that request as well.

What do you think about
```
<rule ref="Squiz.WhiteSpace.OperatorSpacing">
    <properties>
        <property name="ignoreNewlines" value="true" />
    </properties>
</rule>
```
in https://github.com/TYPO3-Documentation/tea/blob/main/phpcs.xml?

jonaseberle:
The previous configuration disallowed newlines in ternary operator like:

```
$a = $condition1 && $condition2
    ? $foo_man_this_is_too_long_what_should_i_do
    : $bar;
```

This seems very arbitrary since breaking on other operators is in fact
allowed.

This change allows such longer statements to be broken apart over
several lines.

Fixes #337

### DIFF
--- a/phpcs.xml
+++ b/phpcs.xml
@@ -122,7 +122,11 @@
     <rule ref="PEAR.WhiteSpace.ScopeClosingBrace"/>
     <rule ref="Squiz.WhiteSpace.CastSpacing"/>
     <rule ref="Squiz.WhiteSpace.LogicalOperatorSpacing"/>
-    <rule ref="Squiz.WhiteSpace.OperatorSpacing"/>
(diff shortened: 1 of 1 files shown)


oliverklee (approved):

oliverklee:
Thanks for contributing, @jonaseberle! :heart:

PR #340, opened 2021-11-17
[TASK] Migrate to typo3 coding standards

lukaszuznanski:
Migrate to typo3 coding standards and php-cs-fixer 3.1, needs #341


oliverklee:
#341 is merged. @lukaszuznanski Can you rebase, please?

.php_cs.php
```diff
@@ -1,186 +0,0 @@
...
-    'void_return' => true,
-    'whitespace_after_comma_in_array' => true,
-];
-
-$config = \PhpCsFixer\Config::create()
-    ->setRiskyAllowed(true)
-    ->setRules($rules);
-
-$finder = \PhpCsFixer\Finder::create()
-    ->in('Classes')
```
oliverklee, line 182: I think we need to keep the directory configuration so we don't check the `vendor` directory.

oliverklee (changes requested):

.php-cs-fixer.php
```diff
@@ -0,0 +1,5 @@
+<?php
+
+$config = \TYPO3\CodingStandards\CsFixerConfig::create();
+$config->getFinder()->in('Classes');
```
oliverklee, line 4: To check the same files as before, we'll also need to check `Tests/` and `Configuration/`.

oliverklee (changes requested):

.php-cs-fixer.php
```diff
@@ -0,0 +1,5 @@
+<?php
```
oliverklee, line 1: We'll also need to update the `.gitattributes` accordingly so the changed file is not exported, and possibly also the pre-release Composer script.

oliverklee (changes requested):

oliverklee (approved):

PR #350, opened 2021-12-23
[TASK] Drop the badge for unstable releases from the README

oliverklee:
The badge behaves in an unpredictable way if there are mutiple unstable
releases (which happens when we update the branch alias for the `main`
branch) and hence is pretty much useless.


oliverklee:
The build failures are unrelated to the changes in the PR. As this PR only contains changes to the README, build failures should not keep us from merging.

lukaszuznanski (approved):

PR #351, opened 2021-12-23
[TASK] Update the development dependencies

oliverklee:

PR #352, opened 2021-12-23
[BUGFIX] Pin Composer to version 2.1.x

oliverklee:
Composer versions 2.2.0 and 2.2.1 have a bug that breaks PHPUnit.
Until version 2.2.2 is released with a fix, we need to keep to
Composer 2.1 to keep the tests from breaking.

See https://github.com/composer/composer/issues/10387 for details.


oliverklee:
The breakage of the composer-normalize step we be fixed by #351.

lukaszuznanski (approved):

lukaszuznanski (approved):

PR #363, opened 2021-12-31
[TASK] Drop support for TYPO3 9LTS

oliverklee:
Now that we have added support for TYPO3 11LTS, we have been supporting
3 different TYPO3 LTS versions in the same branch. This has turned out
to be too much of a hassle when running the test. So now is the time
to drop support for 9LTS.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -140,24 +140,6 @@ jobs:
       fail-fast: false
       matrix:
         include:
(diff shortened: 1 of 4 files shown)


oliverklee:
Let's merge this after we have released version 1.0.1 (as this PR is a breaking change, which needs to happen for version 2).

lukaszuznanski (approved):

PR #370, opened 2022-01-26
[BUGFIX] Avoid xargs error for empty project

tgaertner:
xargs shell command will throw an error if find command will return an empty result. Adding -r will prevent this error. 

Example: 
A project has no yml files.
Before: The yaml-linting fails.
With that patch: The yaml-linting succeed.


oliverklee (approved):
Thanks for contributing this bugfix, @tgaertner! :heart:

PR #373, opened 2022-01-28
[BUGFIX] install and use wait-for-it for functional tests

tgaertner:
It might happen, when starting the fuctional test that the database service container is not ready yet. 

This patch provides an extra timeout to wait for the service to make sure the test does not fail because of missing database.

### DIFF
--- a/.gitlab/build/docker_install.sh
+++ b/.gitlab/build/docker_install.sh
@@ -5,7 +5,7 @@
 set -xe
 
 apt-get update -yqq
-apt-get install git libzip-dev unzip parallel libxml2-utils wget -yqq
+apt-get install git libzip-dev unzip parallel libxml2-utils wget wait-for-it -yqq
 
 php -r "readfile('http://getcomposer.org/installer');" | php -- --install-dir=/usr/local/bin/ --filename=composer
 chmod +x /usr/local/bin/composer
--- a/.gitlab/pipeline/jobs/func-php7.2-v10.yml
+++ b/.gitlab/pipeline/jobs/func-php7.2-v10.yml
@@ -9,4 +9,4 @@ func-php7.2-v10:
     - php-lint-php7.2
   script:
     - composer require --no-progress typo3/minimal:"^10.4"
-    - composer ci:tests:functional
+    - wait-for-it mariadb:3306 -t 60 --strict -- composer ci:tests:functional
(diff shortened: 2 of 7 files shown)


flicstar (commented):
Looks good to me +1
I'll ping someone in the Docs channel to get some more technical eyes on it.

oliverklee:
@flicstar Thanks!

While this project is located under the TYPO3-documentation organization, the QA assurance initiative (#qa-best-practices in Slack) is responsible for it, not the Documentation team,

@lukaszuznanski Would you be a good person to review this change? (I'm not as familiar with GitLab CI as you, I think.)

tgaertner:
While I run into that with GitLab CI it might be worth implementing for other environments, too. Maybe GitHub Pipelines has an internal mechanismen to prevent this or is by design not in danger to run into that, but I'm not familiar with GitHub Pipelines.
What I want to say is, its most likely not a GitLab specific issue - but I implemented it here because this is what I found, what I can reproduce and test.

lukaszuznanski:
> While I run into that with GitLab CI it might be worth implementing for other environments, too. Maybe GitHub Pipelines has an internal mechanismen to prevent this or is by design not in danger to run into that, but I'm not familiar with GitHub Pipelines. What I want to say is, its most likely not a GitLab specific issue - but I implemented it here because this is what I found, what I can reproduce and test.

Could you tell me how to reproduce this issue so I can test it?

tgaertner:
Well, our setup is this:
Selfhosted GitLab CE with Docker-Runner : Docker-Executor

I don't know what really is the key to reproduce this. Maybe its the general performance of the system. It happens even on our environment just 1 out of 3 times that it throws that errors...

The PR just avoid something that can happen. 
The fact that the core and also styleguide has a solution for that, tells me that its no just something special related to our setup.

As pointed out that there netcat is used instead of wait-for-it - so netcat is probably the better/easier solution here?

DanielSiepmann:
We encountered the same issue back in last year.

Not sure about the current state for GitLab CI but I can't find the note in docs any more. They explicitly mentioned that issue back when we had the issue. It looks like they added some auto determinism. But there is still an open bug when multiple ports are part of the service: https://gitlab.com/gitlab-org/gitlab/-/issues/24197

With those new insights: Are you running latest version @tgaertner ? This might be an old issue which was fixed in newer versions via better health check: https://docs.gitlab.com/ee/ci/services/#how-the-health-check-of-services-works

Or as you mentioned performance is to bad and the GitLab timeout of health check is reached. Not sure if the timeout can be configured (I'm not to much into GitLab any more).

But I would say we can accept the workaround, especially as TYPO3 like styleguide already use it as well.

oliverklee (approved):

PR #377, opened 2022-02-09
[TASK] Add composer require checker

### LINKED ISSUE #347: Add composer-require-checker to the toolchain

lukaszuznanski:
Resolves #347


.gitlab/pipeline/jobs/composer-require-checker.yml
```diff
@@ -0,0 +1,5 @@
+composer-normalize:
```
nostadt, line 1: I believe the name needs to be adjusted.

PR #381, opened 2022-02-11
[FEATURE] Enable php8.1

lukaszuznanski:


composer.json
```diff
@@ -25,7 +25,7 @@
 		"source": "https://github.com/TYPO3-Documentation/tea"
 	},
 	"require": {
-		"php": "~7.2.0 || ~7.3.0 || ~7.4.0 || ~8.0",
+		"php": "~7.2.0 || ~7.3.0 || ~7.4.0 || ~8.0 || ~8.1",
```
oliverklee, line 28: We'll also need to update the `ext_emconf.php`.

oliverklee (changes requested):

oliverklee:
And we need to switch to the core testing framework first. (The nimut TF doesn't support PHP 8.1 yet.)

.github/workflows/ci.yml
```diff
@@ -220,12 +227,40 @@ jobs:
...
+            composer-dependencies: lowest
+          - typo3-version: ^11.5
+            php-version: 8.0
+            composer-dependencies: highest
+          - typo3-version: ^11.5
+            php-version: 8.0
+            composer-dependencies: lowest
+          - typo3-version: ^11.5
+            php-version: 8.1
+            composer-dependencies: highest
```
oliverklee, line 263: We need to leave out the functional tests for V11 until we have fixed them (in a separate PR).

oliverklee (changes requested):

oliverklee (approved):

PR #396, opened 2022-02-24
[BUGFIX] Use the correct Composer cache directory for CI caching

oliverklee:
The cache directory has changed with Composer 2, and we need to
adapt our CI cache configuration accordingly.


oliverklee:
Closing. We are caching the correct directory.

PR #405, opened 2022-03-22
[BUGFIX] Fix Gitab  pipeline

lukaszuznanski:


composer.json
```diff
@@ -34,7 +34,7 @@
 	"require-dev": {
 		"codeception/codeception": "^4.1.30",
 		"doctrine/dbal": "^2.13.8",
-		"ergebnis/composer-normalize": "^2.19.0",
+		"ergebnis/composer-normalize": "2.18.0",
```
oliverklee, line 37: Why pin the version?

oliverklee:
My informed guess about the failure is that GitLab CI caches the vendor folder inbetween builds (instead of the Composer cache folder), keeping Composer from downgrading Composer Normalize on builds with PHP versions that require a lower version.

So we should cache the Composer cache director, not the vendor folder, and use the PHP version, TYPO3 version, hashed `composer.json` and highest/lowest settings as cache keys.

PR #417, opened 2022-04-06
[TASK] Bump the code coverage generation to TYPO3 11LTS and PHP 8.1

oliverklee:
We should use the latest TYPO3 and PHP versions for code coverage
generation.


lukaszuznanski (approved):

lukaszuznanski:
Possibly, but it's very hard to test without merging (requires changes to code coverage generation)

PR #445, opened 2022-04-29
[TASK] Move npm tools and config to default places

### LINKED ISSUE #444: Move npm tools and config to default locations
Most IDEs pickup the configuration for QA tools automatically from default locations. This eases adaption as it uses zero-config and zero-knowledge as the IDE option "format code" picks up the correct config.

While having the config in a separate location might "look less cluttered", it still is a benefit if a IDE button does automatically what a defined CI/composer script would do. So instead of creating the need to configure the IDE with the special config path, or enforce using a composer script to run tools, the config should be available in the default spot for better discoverability.

The benefits of course are only valid if a modern IDE with suitable plugins is used, however the downside of "cluttered root folder" can be easily circumvented by IDE plugins and views which hide or fold away such default files. Additionally while using dot-prefixed-files to more or less auto-hide them is sometimes an option, not all tools allow for such an autodetected alternative.

jpmschuler:
Most IDEs pick up the configuration for QA tools automatically from default locations. This eases adaption as it uses zero-config and zero-knowledge as the IDE option "format code" picks up the correct config.

While having the config in a separate location might "look less cluttered", it still is a benefit if a IDE button does automatically what a defined CI/Composer script would do. So instead of creating the need to configure the IDE with the special config path, or enforce using a Composer script to run tools, the config should be available in the default spot for better discoverability.

Fixes #444


jpmschuler:
`lint:js` disrespects `.editorconfig` now, looking into it

.gitattributes
```diff
@@ -7,11 +7,11 @@
 /.gitlab/ export-ignore
 /.php-cs-fixer.php export-ignore
 /.phpstorm.meta.php export-ignore
-/Resources/Private/.eslintignore export-ignore
-/Resources/Private/.eslintrc.json export-ignore
-/Resources/Private/.prettierrc.js export-ignore
-/Resources/Private/package.json export-ignore
-/Resources/Private/stylelint.config.js export-ignore
+/.eslintignore export-ignore
```
oliverklee, line 10: Please sort the lines in this file. In PhpStorm: Edit -> Sort Lines

.gitignore
```diff
@@ -3,10 +3,10 @@
...
 /Documentation-GENERATED-temp/
-/Resources/Private/node_modules/
-/Resources/Private/package-lock.json
-/Resources/Private/yarn-error.log
-/Resources/Private/yarn.lock
+/node_modules/
+/package-lock.json
+/yarn-error.log
+/yarn.lock
 /build
```
oliverklee, line 10: Ditto about keeping the lines sorted.

.editorconfig
```diff
@@ -10,10 +10,6 @@ indent_size = 4
 insert_final_newline = true
 trim_trailing_whitespace = true
 
-# JS files
-[*.js]
```
oliverklee, line 14: We should keep this in sync with https://github.com/TYPO3/coding-standards/blob/main/.editorconfig, where 2 spaces are used for indenting JS files. (Also, we should keep `.js` files in the `.editorconfig` so PhpStorm can use this for autormatting.)

oliverklee (changes requested):

oliverklee (changes requested):
Please also add a line to the changelog.

oliverklee (changes requested):
Please also adapt the `ci:json:lint` Composer command to check the moved `package.json`, but not the moved `node_modules/`.

jpmschuler:
Done, there where a few more composer scripts adaptions necessary (prepare-release and c:yaml:lint)

oliverklee (approved):
Looks good to me, but I'd like to have another pair of eyes on this. @tgaertner Would you do the second review?

lukaszuznanski (changes requested):
I think that pipeline for gitlab requires changes here
https://github.com/TYPO3-Documentation/tea/blob/main/.gitlab/pipeline/jobs/.default-frontend.yml#L11

jpmschuler:
Adapted gitlab script and tested in https://gitlab.com/jpmschuler/tea/-/pipelines/542194787 that 

- ts:lint https://gitlab.com/jpmschuler/tea/-/jobs/2473820989 and
- style:lint https://gitlab.com/jpmschuler/tea/-/jobs/2473820974 work

CHANGELOG.md
```diff
@@ -10,6 +10,7 @@ This project adheres to [Semantic Versioning](https://semver.org/).
 
 ### Changed
 - Use the TYPO3 Code of Conduct (#430)
+- Move npm tools and config to default locations (#444)
```
oliverklee, line 13: This needs to move to the top of the list (as the changelog is from newest to oldest).

oliverklee (changes requested):

.gitlab/pipeline/jobs/.default-frontend.yml
```diff
@@ -7,6 +7,4 @@
       - Resources/Private/node_modules/
       - Resources/Private/.yarn
   before_script:
-    - npm install -g eslint
```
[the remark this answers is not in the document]
jpmschuler: This command installs ESLINT globally, however the yarn command next to it install ESLINT locally (as that package is a dependency), thus the global install is neither necessary, nor helpful (as it doesn't define a version and thus could load a new major with breaking changes or similar).
The Github action doesn't have that command.

CHANGELOG.md
jpmschuler: done

lukaszuznanski (approved):

PR #459, opened 2022-06-15
[BUGFIX] Treat PHP versions in the GitHub CI configuration as strings

oliverklee:
PHP version numbers need to be strings: If the version `8.0` were
treated as a number, the YAML parser would happily convert it to `8`,
hence dropping the minor version number.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -28,11 +28,11 @@ jobs:
       fail-fast: false
       matrix:
         php-version:
-          - 7.2
-          - 7.3
-          - 7.4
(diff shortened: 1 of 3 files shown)


tgaertner (changes requested):
What about other version strings like typo3-version. I guess there is no problem right now because of the "^" in front, but if this get changed for some reason the same problem rises up - right? So to make that save and streamline with php-version we might add the "" forcing string here, too?

oliverklee:
Done.

tgaertner (approved):
Looks good to me. Pipeline is green

PR #461, opened 2022-06-25
[FEATURE] Configure Dependabot to update Composer dev dependencies

oliverklee:


tgaertner (approved):
Well I'm not really into dependabot or github actions. 
Code looks good to me and checks have passed, so this should be ok.

PR #471, opened 2022-06-28
[TASK] Add the PHPStan strict rules

oliverklee:
We want PHPStan to be as strict as possible.

(The new warning requires a change in Extbase to get solved properly.)


oliverklee:
This is the Core fix: https://review.typo3.org/c/Packages/TYPO3.CMS/+/74991

lukaszuznanski (approved):

Issue #475, opened 2022-07-05
Better eslint configuration for backend JavaScript


mabolek:
When I work with JavaScript for the backend, this `.estlintrc.json` config works better. Eslint will no longer complain about `TYPO3`, `$` and `define()` calls, etc. It will also allow me to use `const`.

``` json
{
	"root": true,
	"extends": [
		"eslint:recommended",
		"plugin:prettier/recommended"
	],
	"env": {
		"browser": true,
		"node": true,
		"es6": true,
		"amd": true,
		"jquery": true
	},
	"globals": {
		"TYPO3": "readonly"
	},
	"parserOptions": {
		"ecmaVersion": 8
	}
}
```

mabolek:
@oliverklee ⬆️

PR #476, opened 2022-07-05
[TASK] Improved ESLint config for Backend JavaScript

mabolek:
With this configuration, ESLint will no longer complain about the `TYPO3` and `$` global, `define()` calls, etc. It will also allow the use of `const`.

Resolves: #475

### DIFF
--- a/.eslintrc.json
+++ b/.eslintrc.json
@@ -5,6 +5,16 @@
 		"plugin:prettier/recommended"
 	],
 	"env": {
-		"browser": true
+		"browser": true,
+		"node": true,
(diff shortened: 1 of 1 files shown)


oliverklee:
I'm not that familiar with eslint and hence am hesitant reviewing this change. @lukaszuznanski Would this fall in your area of expertise?

lukaszuznanski:
Not really, I think it's better to ask some guy which has more javascript experience.

michsch (approved):

oliverklee (approved):

PR #494, opened 2022-07-26
V2.0.0: Drop support for TYPO3 9LTS

oliverklee:


oliverklee:
A note about the unusual commit message: For releases, the commit message will be the release message in the TER. So it needs to serve that purpose.

tgaertner:
Changes looks ok for me.

tgaertner (approved):

PR #503, opened 2022-08-12
Add links to CHANGELOG.md

sypets:
Convert the PR number references to links.

Resolves: #503

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -18,58 +18,58 @@ This project adheres to [Semantic Versioning](https://semver.org/).
 ## 2.0.0
 
 ### Added
-- Add the PHPStan strict rules (#471)
-- Add a Dependabot action for updating dependencies (#452, #461, #469, #481)
-- Use Coveralls for the code coverage (#425)
+- Add the PHPStan strict rules ([#471](#471))
+- Add a Dependabot action for updating dependencies ([#452](#452), [#461](#461), [#469](#469), [#481](#481))
+- Use Coveralls for the code coverage ([#425](#425))
 
 ### Changed
-- Move npm tools and config to default locations (#444)
-- Use the TYPO3 Code of Conduct (#430)
+- Move npm tools and config to default locations ([#444](#444))
+- Use the TYPO3 Code of Conduct ([#430](#430))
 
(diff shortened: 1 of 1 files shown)


CHANGELOG.md
```diff
@@ -18,68 +18,68 @@ This project adheres to [Semantic Versioning](https://semver.org/).
 ## 2.0.0
 
 ### Added
-- Add the PHPStan strict rules (#471)
-- Add a Dependabot action for updating dependencies (#452, #461, #469, #481)
-- Use Coveralls for the code coverage (#425)
+- Add the PHPStan strict rules ([#471](#471))
```
oliverklee, line 21: This doesn't seem to work - the links now are anchor links instead of getting expand to issue/PR links:
https://github.com/sypets/tea/blob/add-links-to-changelog/CHANGELOG.md

oliverklee (changes requested):

CHANGELOG.md
oliverklee: We might need to link the full URLs there (which I'd like to avoid as it is quite a hassle and (as proven in another project) very error prone): https://raw.githubusercontent.com/sebastianbergmann/phpunit/main/ChangeLog-10.0.md

oliverklee: This is the project where we switch for these exact reasons: https://github.com/MyIntervals/emogrifier/pull/976

oliverklee:
NB: In the release notes, the autolinking of the issues/PRs works fine: https://github.com/TYPO3-Documentation/tea/releases/tag/v2.0.0

CHANGELOG.md
sypets: > This doesn't seem to work - the links now are anchor links instead of getting expand to issue/PR links

Right. Sorry about that. So used to using the issue number in commit messages etc.

I agree the full URLs are not such a good solution. If this is too much of a hassle we can abandon.

Using the release notes might be an option. I notice several extensions authors do this. If this extension is used to "standardize" things it would be important to use something "hassle-free" which is not much extra work.

oliverklee:
@sypets Thanks for digging into this!

PR #511, opened 2022-08-23
[BUGFIX] Explicitly require Prophecy

oliverklee:
PHPUnit has dropped its dependency on Prophecy. In order to keep using
Prophecy in our tests, we need to add it as an explicit (development)
dependency.

Also, relying on transitive dependencies is bad practice anyway.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -15,6 +15,7 @@ This project adheres to [Semantic Versioning](https://semver.org/).
 ### Removed
 
 ### Fixed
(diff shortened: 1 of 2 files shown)


oliverklee:
This PR blocks all other current PRs.

Note: The test failures are unrelated. I'll create a separate PR for that. (We'll need to merge at least one of the two blocker PRs even with a red pipeline.)

lukaszuznanski (approved):

PR #512, opened 2022-08-23
[TASK] Remove the ancient acceptance tests

oliverklee:
The acceptance tests have not been testing the extension itself,
have not been maintained for quite some time, and do not serve
as a good example anymore.

Until we have a proper set of example acceptance tests, we should
remove them.

Also drop the now-unused dependency on Codeception.


oliverklee:

brotkrueml (approved):

PR #513, opened 2022-08-23
[TASK] Update to Composer 2.4

oliverklee:
Now that Composer 2.4 has had its first point release, we can safely
upgrade from 2.3 to 2.4 in the DDEV configuration and the CI pipeline.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -8,6 +8,7 @@ This project adheres to [Semantic Versioning](https://semver.org/).
 ### Added
 
 ### Changed
+- Update to Composer 2.4 (#513)
 - Change the default indentation for rst files to 4 spaces (#194)
 
 ### Deprecated
--- a/.ddev/config.yaml
+++ b/.ddev/config.yaml
@@ -18,7 +18,7 @@ hooks:
 omit_containers: [dba, ddev-ssh-agent]
 webimage_extra_packages: [parallel]
 use_dns_when_possible: true
-composer_version: "2.3.7"
+composer_version: "2.4"
 web_environment:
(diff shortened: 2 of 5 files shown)


oliverklee:

brotkrueml (commented):
What is the background behind defining the composer version in GitHub actions? The most recent version is used when nothing is defined. Or do you want to update to a new minor version explicitly to avoid possible breakings (which should not occur in minor versions)?

Also, as a cron is configured, this would give early response after that run when an new composer version breaks something. So, maybe "v2" is enough?

oliverklee:
Indeed, the reason is that minor versions sometimes have (for us) breaking changes. So I'd like to keep `v2.4` (instead of just `v2`). We don't need the patch-level release anymore, though. I'll update the PR in a second.

oliverklee:
Done and repushed.

CHANGELOG.md
```diff
@@ -8,6 +8,7 @@ This project adheres to [Semantic Versioning](https://semver.org/).
 ### Added
 
 ### Changed
+- Upgrade to Composer 2.4 (#513)
```
brotkrueml, line 11:
```suggestion
- Update to Composer 2.4 (#513)
```

Update -> minor/patch level version
Upgrade -> major version

brotkrueml (changes requested):

CHANGELOG.md
oliverklee: Done.

brotkrueml (approved):

brotkrueml:
Seems that I don't have the rights to approve it to be unblocking. The approval is only grey, not green.

PR #515, opened 2022-08-23
[BUGFIX] Have all extension dependencies in the `ext_emconf.php` as well

oliverklee:
For non-Composer installations, all extension dependencies that are in
the `composer.json` should also be in the `ext_emconf.php`.

Even for extensions that are always installed (like frontend or fluid),
it is good practice to always make these dependencies explicit.


oliverklee:

lukaszuznanski (approved):

PR #521, opened 2022-08-28
[TASK] Substitute TYPO3_MODE constant with TYPO3 constant

brotkrueml:
The TYPO3_MODE constant is deprecated with v11.0. The new constant
TYPO3, which replaces the old constant, was introduced also in
TYPO3 v10 with version v10.4.11:

https://docs.typo3.org/c/typo3/cms-core/main/en-us/Changelog/11.0/Deprecation-92947-DeprecateTYPO3_MODEAndTYPO3_REQUESTTYPEConstants.html

As best practise, we should also use the new constant and
raise the composer constraints for TYPO3 accordingly.


linawolf (approved):

composer.json
```diff
@@ -28,10 +28,10 @@
 	},
 	"require": {
 		"php": "~7.2.0 || ~7.3.0 || ~7.4.0 || ~8.0.0 || ~8.1.0",
-		"typo3/cms-core": "^10.4 || ^11.5.2",
-		"typo3/cms-extbase": "^10.4.6 || ^11.5.2",
-		"typo3/cms-fluid": "^10.4 || ^11.5.2",
-		"typo3/cms-frontend": "^10.4 || ^11.5.2"
+		"typo3/cms-core": "^10.4.11 || ^11.5.2",
```
oliverklee, line 31: It should suffice to raise the version of core - the other extension can stay at their minimum requirements.

oliverklee (changes requested):

composer.json
linawolf: Does that really make sense? You cannot use a different version of typo3/cms-core and typo3/cms-extbase anyway. And then why for example do we also require certain bug-fix versions for the other system extensions for TYPO3 v11?

oliverklee: The V11 thing indeed is an inconsistency. I’d like to discuss our policy concerning how to handle system extension versions at our next call (of the best practices team).

oliverklee:
Marking as blocked until we have decided on how to proceed with the version numbers (unless @brotkrueml you would be willing to only raise core to get this merged).

oliverklee (approved):

Issue #546, opened 2022-09-01
`composer ci:php:lint` ignores hidden files


oliverklee:
It should also find hidden files, e.g. `.phpstorm.meta.php`.

We probably need to add `.*.php` to the `find`.

PR #559, opened 2022-09-06
[BUGFIX] Add hidden files to find in PHP lint

lukaszuznanski:
Resolves: #546


oliverklee:
It looks like our PhpStorm meta file is not compatible with PHP 7.2.

oliverklee (approved):

Issue #542, opened 2022-09-01
Upgrade to PHPCOV 8


oliverklee:
This needs PHPUnit 9.

PR #563, opened 2022-09-11
Update phpunit to v9

lukaszuznanski:
Update to PHPUnit version 9 in order to unblock #542. Also adds prophesize-phpunit and rewrites deprecation.


lukaszuznanski:
Well, it looks like we have a problem. Prophesize-phpunit provides Prophesize trait only in version > 2, and in this version, only php 7.3 > is supported. Current approach throws warning, because it's deprecated.

oliverklee:
We also need to drop support for PHP 7.2 and 7.3 first. (Hence moving to the 3.0 milestone.)

lukaszuznanski:
@oliverklee I will close this as i can see that #548 actually solves this.

PR #573, opened 2022-09-20
[TASK] Upgrade the functional tests on CI to Ubuntu 22.04

### LINKED ISSUE #173: Get the functional tests to run with the ubuntu-22.04 runner
Currently, ubuntu-18.04 works fine, but ubuntu-20.04 does not. Maybe we can adopt https://github.com/oliverklee/ext-phpunit/blob/main/.github/workflows/ci.yml for this.

GItHub now will have weekly brownouts with the 18.04 runners: https://github.com/actions/runner-images/issues/6002

oliverklee:
Fixes #173

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -182,8 +182,13 @@ jobs:
             composer-dependencies: lowest
   functional-tests:
     name: "Functional tests"
-    runs-on: ubuntu-18.04
+    runs-on: ubuntu-22.04
     needs: php-lint
+    env:
+      DB_DATABASE: typo3
+      DB_USER: root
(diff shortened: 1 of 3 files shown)


oliverklee:
It looks like this needs to wait until we have dropped support for PHP 7.2 and 7.3.

oliverklee:
I'll need to have a look into this bug: https://github.com/phpmyadmin/phpmyadmin/issues/14220

oliverklee:
Yee-haw, I have been able to solve the PHP 7.2/7.3 problem!

lukaszuznanski (approved):

PR #581, opened 2022-09-21
[TASK] Drop support for PHP 7.2 and 7.3

### LINKED ISSUE #577: Drop the PHP 7.2/7.3 workaround for MySQL from the CI scripts
… once we have dropped support for PHP 7.2 and 7.3.

oliverklee:
Also sync the test matrices of the `predefined` CI jobs with those from the Composer-script-driven CI jobs.

This change does not add any 7.4-only PHP language features.

Fixes #577


CHANGELOG.md
```diff
@@ -12,6 +12,7 @@ This project adheres to [Semantic Versioning](https://semver.org/).
 ### Deprecated
 
 ### Removed
+- Drop support for PHP 7.2 and 7.3 (#581)
```
brotkrueml, line 15:
```suggestion
- Support for PHP 7.2 and 7.3 (#581)
```

It is read: "Removed support for ...", so I suggest to avoid "Drop".

brotkrueml (commented):
I assume, phpunit and maybe other dependencies are updated in a follow-up PR?

oliverklee:
> I assume, phpunit and maybe other dependencies are updated in a follow-up PR?

Correct. I'd like to keep the PRs small and focused ("only one topic per PR").

CHANGELOG.md
oliverklee: It'd prefer the changelog lines to always be worded like good commit message subjects. i.e., starting with a imperative verb like explained here: https://cbea.ms/git-commit/

And the commit message entries should make sense on its own, i.e., they should work without having to add the section heading. (This also is how we have tried to write the changelog entries so far.)

brotkrueml (approved):

Issue #564, opened 2022-09-11
Return ResponseInterface in Controller Actions


linawolf:
With version 11 a controller action should return a ResponseInterface, this becomes mandatory with v12

See https://docs.typo3.org/c/typo3/cms-core/main/en-us/Changelog/11.0/Deprecation-92784-ExtbaseControllerActionsMustReturnResponseInterface.html

PR #597, opened 2022-09-28
[TASK] Return `ResponseInterface` in controller actions

linawolf:
resolves https://github.com/TYPO3-Documentation/tea/issues/564

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -9,6 +9,7 @@ This project adheres to [Semantic Versioning](https://semver.org/).
 
 ### Changed
 - Replace switchable controller actions with separate plugins (#575)
+- Controller actions return ResponseInterface (#597)
 
 ### Deprecated
 
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -8,7 +8,7 @@ This project adheres to [Semantic Versioning](https://semver.org/).
 ### Added
 
 ### Changed
-- Return ResponseInterface in controller actions (#597)
+- Return `ResponseInterface` in controller actions (#597)
 - Replace switchable controller actions with separate plugins (#575)
 
 ### Deprecated
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -8,8 +8,8 @@ This project adheres to [Semantic Versioning](https://semver.org/).
 ### Added
 
 ### Changed
+- Return ResponseInterface in controller actions (#597)
 - Replace switchable controller actions with separate plugins (#575)
-- Controller actions return ResponseInterface (#597)
 
 ### Deprecated
 
--- a/Classes/Controller/TeaController.php
+++ b/Classes/Controller/TeaController.php
@@ -19,10 +19,9 @@ class TeaController extends ActionController
      */
     private $teaRepository;
 
-    public function injectTeaRepository(TeaRepository $teaRepository): ResponseInterface
+    public function injectTeaRepository(TeaRepository $teaRepository): void
     {
         $this->teaRepository = $teaRepository;
-        return $this->htmlResponse();
     }
 
     public function indexAction(): ResponseInterface
--- a/Classes/Controller/TeaController.php
+++ b/Classes/Controller/TeaController.php
@@ -4,6 +4,7 @@ declare(strict_types=1);
 
 namespace TTN\Tea\Controller;
 
+use Psr\Http\Message\ResponseInterface;
 use TTN\Tea\Domain\Model\Product\Tea;
 use TTN\Tea\Domain\Repository\Product\TeaRepository;
 use TYPO3\CMS\Extbase\Mvc\Controller\ActionController;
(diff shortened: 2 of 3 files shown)


Classes/Controller/TeaController.php
```diff
@@ -18,18 +19,21 @@ class TeaController extends ActionController
      */
     private $teaRepository;
 
-    public function injectTeaRepository(TeaRepository $teaRepository): void
+    public function injectTeaRepository(TeaRepository $teaRepository): ResponseInterface
```
brotkrueml, line 22: The inject method does not return anything, it is for injecting a dependency. Is there a reason not using constructor injection?

DavidBruchmann: the inject method never has to return a ResponseInterface I suppose.
can be the old value void
[3:05](https://typo3.slack.com/archives/CSQ8T0VS5/p1664395534686369)
only actions return a ResponseInterface afaik

sbuerk: inject methods should be return type void and not returning anything. Does not make any sense anyway.

Tests/Unit/Controller/TeaControllerTest.php
```diff
@@ -51,6 +58,11 @@ protected function setUp(): void
         /** @var TeaRepository&ProphecySubjectInterface $teaRepository */
         $teaRepository = $this->teaRepositoryProphecy->reveal();
         $this->subject->injectTeaRepository($teaRepository);
+
+        $this->responseFactoryProphecy = $this->prophesize(ResponseFactory::class);
+        /** @var ResponseFactory&ProphecySubjectInterface $responseFactory */
+        $responseFactory = $this->responseFactoryProphecy->reveal();
```
sbuerk, line 64: `$this->subject->injectTeaRepositoy()` is changed to return `$this->htmlResponse()` .. which is wrong anyway. You added code to inject the responseFactory .. but "to late". The responseFactory must be existing in the actionController before first usage (which is added in `injectTeaRepository` ...

Tests/Unit/Controller/TeaControllerTest.php
```diff
@@ -69,6 +81,7 @@ public function indexActionAssignsAllTeaAsTeasToView(): void
         $teas = $this->prophesize(QueryResultInterface::class)->reveal();
         $this->teaRepositoryProphecy->findAll()->willReturn($teas);
         $this->viewProphecy->assign('teas', $teas)->shouldBeCalled();
+        $this->responseFactoryProphecy->createResponse()->shouldBeCalled();
```
sbuerk, line 84: Eventually add here a `->willReturn(new HtmlResponse())` or a proper prophesized response or setup response. Eventually do not prophecies the response factory if possibile (not looked into it)

Classes/Controller/TeaController.php
linawolf: yeah, I was kind of blind

Tests/Unit/Controller/TeaControllerTest.php
linawolf: I didn't find a way not to prophesize this and still get the tests running However it feels like I am testing Core Details now and not really the Tea Controller...

CHANGELOG.md
```diff
@@ -9,6 +9,7 @@ This project adheres to [Semantic Versioning](https://semver.org/).
 
 ### Changed
 - Replace switchable controller actions with separate plugins (#575)
+- Controller actions return ResponseInterface (#597)
```
oliverklee, line 12: Please move the entry to the top of the section so the changes are in reverse chronological order ("newest on top").

Please use the same format in the changelog as in the commit message subject, i.e., use the imperative:

Return `ResponseInterface` in controller actions

For details on improving git commit message, I can highly recommend this article: https://cbea.ms/git-commit/

Tests/Unit/Controller/TeaControllerTest.php
```diff
@@ -51,6 +65,16 @@ protected function setUp(): void
...
+
+        $this->responseFactoryProphecy = $this->prophesize(ResponseFactory::class);
+        /** @var ResponseFactory&ProphecySubjectInterface $responseFactory */
+        $responseFactory = $this->responseFactoryProphecy->reveal();
+        $this->subject->injectResponseFactory($responseFactory);
+
+        $this->streamFactoryProphecy = $this->prophesize(StreamFactory::class);
+        /** @var StreamFactory&ProphecySubjectInterface $streamFactory */
+        $streamFactory = $this->streamFactoryProphecy->reveal();
+        $this->subject->injectStreamFactory($streamFactory);
```
oliverklee, line 77: You can simplify this a lot by adding `htmlResponse` to the list of mocked methods in line 58.

Classes/Controller/TeaController.php
```diff
@@ -23,13 +24,15 @@ public function injectTeaRepository(TeaRepository $teaRepository): void
         $this->teaRepository = $teaRepository;
     }
 
-    public function indexAction(): void
+    public function indexAction(): ResponseInterface
```
oliverklee, line 27: We should also have tests that the actions return a `HtmlResponse` with the content from `htmlResponse`.

oliverklee (changes requested):

CHANGELOG.md
```diff
@@ -8,8 +8,8 @@ This project adheres to [Semantic Versioning](https://semver.org/).
 ### Added
 
 ### Changed
+- Return ResponseInterface in controller actions (#597)
```
oliverklee, line 11: And to make the changelog even more beautiful, please put code, class names etc. (in this case, `ResponseInterface`) in backticks so it gets rendered in a monospace font. (GitHub parses Markdown in the release notes, too.)

Tests/Unit/Controller/TeaControllerTest.php
linawolf: Done

Classes/Controller/TeaController.php
linawolf: Done

oliverklee (approved):
LGTM.

PR #620, opened 2022-10-09
[TASK] Remove obsolete Composer conflicts

oliverklee:
The removed conflicting versions are not possible anymore with our current set of dependencies.

### DIFF
--- a/composer.json
+++ b/composer.json
@@ -60,7 +60,6 @@
 		"typo3-ter/tea": "self.version"
 	},
 	"conflict": {
-		"doctrine/dbal": "2.13.1",
 		"typo3/class-alias-loader": "< 1.1.0"
 	},
 	"prefer-stable": true,


composer.json
```diff
@@ -60,9 +60,7 @@
 		"typo3-ter/tea": "self.version"
 	},
 	"conflict": {
-		"doctrine/dbal": "2.13.1",
-		"friendsofphp/php-cs-fixer": "3.5.0",
-		"typo3/class-alias-loader": "< 1.1.0"
```
brotkrueml, line 65: For the minimum required TYPO3 version 11.5.2 the class-alias-loader may be 1.0:
Definition in composer.json of TYPO3 11.5.2:

    "typo3/class-alias-loader": "^1.0"

oliverklee: Oh, right. Done.

brotkrueml (approved):

PR #625, opened 2022-10-09
[TASK] Make the TCA ready for TYPO3 v12

### LINKED ISSUE #624: Make the TCA ready for TYPO3 v12

linawolf:
resolves #624


oliverklee (changes requested):
We'll also need to update the file upload. See this build failure for details (11.2, 8.1, highest):
https://github.com/TYPO3-Documentation/tea/actions/runs/3214806054/jobs/5255454647

linawolf:
Yeah right, I just wrote the docu for that ;)

Configuration/TCA/Overrides/tx_tea_domain_model_product_tea.php
```diff
@@ -0,0 +1,26 @@
+<?php
+
+defined('TYPO3') || die();
+
+$typo3Version = new \TYPO3\CMS\Core\Information\Typo3Version();
+
+if ($typo3Version->getMajorVersion() < 12) {
```
oliverklee, line 7: I propose moving this from the overrides directly to `Configuration/TCA/tx_tea_domain_model_product_tea.php`.

brotkrueml (commented):
I don't get it: tea main branch is by now compatible with 11.5, so why make this construct with the if < 12 condition?

When main should be made compatible with both, 11.5 and 12.0, a TCA migration for eval's is in place by core, so there is no need for this now.

cruser_id is ignored in v12. So there is also no need for omitting this.

Also: "required" is possibly also in 11.5 as separate key, not only in eval. This can be migrated now.

linawolf:
@brotkrueml This is a commit for the Milestone to add TYPO3 v12 support.

Configuration/TCA/Overrides/tx_tea_domain_model_product_tea.php
linawolf: The files contained directly TCA folder should contain only a direct return of the TCA array. News is for example also solving it like that: https://github.com/georgringer/news/blob/main/Configuration/TCA/Overrides/tx_news_domain_model_news.php

oliverklee: > The files contained directly TCA folder should contain only a direct return of the TCA array.

Then let's do that. Or am I missing something here?

Usually, for version-specific switches, I do it like that in my extensions:

```php
$tca = […];
// modify $tca depending on which TYPO3 version we're on
return $tca;
```

linawolf: I think as long as you don't do anything that mustn't be cached that sounds fine - and versions shouldn't change within the cache lifetime

linawolf:
@oliverklee What else is needed to merge this PR?

PR #639, opened 2022-10-16
[TASK] Postpone the TYPO3 12 compatibility

oliverklee:
We should wait with marking this extension TYPO3-12-compatible until an installation with TYPO3 12 is actually possible (i.e., until all development dependencies actually allow this), as an installation with TYPO3 12 will require changes to the PHPStan annotations.

Otherwise, we will get PHPStan errors for the first build after all dependencies allow an installation with TYPO3 12.

This reverts commit 639e41451212ee8b92c0bdf1a927068e25f2ced2.


brotkrueml (commented):
Why not revoke the initial commit?

oliverklee:
Thai’s what I’m doing here: 

> This reverts commit https://github.com/TYPO3-Documentation/tea/commit/639e41451212ee8b92c0bdf1a927068e25f2ced2.

brotkrueml (approved):

Issue #614, opened 2022-10-07
Harden the GitHub Actions workflows


oliverklee:
Let's check whether we should do what https://github.com/vimeo/psalm/pull/8451/files does.

lukaszuznanski:
I can see that those directives applies to `GITHUB_TOKEN`. It's used in code coverage workflow in tea repository. I'm willing to provide PR.

PR #649, opened 2022-10-18
Harden Github Actions workflows

lukaszuznanski:
Resolves: #614


oliverklee (approved):
Looks good. Could you please rebase?

oliverklee (changes requested):
We also should reduce permissions in all other workflows to `contents: read`.

oliverklee (approved):

Issue #493, opened 2022-07-22
Use the PHPStan GitHub error format in CI


oliverklee:
Like this: https://github.com/TYPO3/styleguide/pull/336

sbuerk:
To get this work, https://github.com/TYPO3-Continuous-Integration/TYPO3-CI-PHPStan must add a additional configuration option to provide additional phpstan options. So that it can be set in the github action workflow.

Other way round would be to not use that github workflow/action and directly install/call phpstan.

lukaszuznanski:
> To get this work, https://github.com/TYPO3-Continuous-Integration/TYPO3-CI-PHPStan must add a additional configuration option to provide additional phpstan options. So that it can be set in the github action workflow.
> 
> Other way round would be to not use that github workflow/action and directly install/call phpstan.
> 
> ```yaml
>   phpstan-check:
>     name: "PHPStan check"
>     runs-on: ubuntu-22.04
>     needs: php-lint
>     steps:
>       - name: "Checkout"
>         uses: actions/checkout@v3
>       - name: "Run PHPStan"
>         uses: TYPO3-Continuous-Integration/TYPO3-CI-PHPStan@v1
>         with:
>           config_file: './phpstan.neon'
>           # [1] provide phpstan error format
>           error-format: 'github'
>           # [2] or in a more generic way
>           options: '--error-format=github"'
> ```

TYPO3-CI-PHPStan is only used in one of our 2 main pipelines. I will provide PR for TYPO3-CI-PHPStan, but this doesn't resolve whole problem. Actually in github actions ci.yaml we are using matrix of commands, so maybe the best way is to create separate composer script with this additional parameter? Because as I understand, this parameter doesn't provide any value when set on local environment, is it true?

sbuerk:
The error format is only for the phpstan tool. So the matrix of other jobs are not relevant here.

Generally there are 3 options here for this task, as far as I can see:

* extend `TYPO3-Continuous-Integration/TYPO3-CI-PHPStan` with option for error format and use it then
* drop `TYPO3-Continuous-Integration/TYPO3-CI-PHPStan` and directly call the phpstan binard with needed commands in the github action
* drop `TYPO3-Continuous-Integration/TYPO3-CI-PHPStan`, add an additional command with github error reporting format and call composer script from github action `"ci:php:stan:github"` next to `"ci:php:stan"` ....

PR #663, opened 2022-10-25
[TASK] Use GitHub error formatting for phpstan CI

lukaszuznanski:
Resolves: #493

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -72,6 +72,7 @@ jobs:
           - "json:lint"
           - "php:cs-fixer"
           - "php:sniff"
+          - "php:stan"
           - "php:stan:github"
           - "ts:lint"
           - "yaml:lint"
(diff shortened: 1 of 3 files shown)


oliverklee:
I have tested this [in a PR](https://github.com/TYPO3-Documentation/tea/pull/666) and would expect to be able to see something there. Am I missing something?

lukaszuznanski:
Output should be different. I will provide separate pr with bug that phpstan should catch and we will compare them.

oliverklee:
> Using this new command shows new output in CLI, which I believe is for github actions, but in github actions I can't see any difference. Maybe I'm missing something here?

I honestly have no idea. Maybe we need to do some more research on this.

lukaszuznanski:
@oliverklee I think it's enabled by default now, take a look here:
https://github.com/FriendsOfTYPO3/tea/pull/889/files
it looks like it's enabled, I think we can close this PR

oliverklee:
How can I see whether it's enabled?

lukaszuznanski:
It was visible in this PR when phpstan failed, it's not anymore :/

lukaszuznanski:
Closing for now, will reopen with clean branch

PR #685, opened 2022-11-21
[BUGFIX] Use plural in the TypoScript template settings

### LINKED ISSUE #684: Use plural in the TypoScript template settings
Instead of having a simple `templateRootPath` etc., we need to use the plural as described on https://docs.typo3.org/m/typo3/book-extbasefluid/10.4/en-us/b-ExtbaseReference/Index.html.

lukaszuznanski:
Resolves #684

### DIFF
--- a/Configuration/TypoScript/constants.typoscript
+++ b/Configuration/TypoScript/constants.typoscript
@@ -1,13 +1,4 @@
 plugin.tx_tea {
-  view {
-    # cat=plugin.tx_tea/file; type=string; label=Path to template root (FE)
-    templateRootPaths = EXT:tea/Resources/Private/Templates/
-    # cat=plugin.tx_tea/file; type=string; label=Path to template partials (FE)
-    partialRootPaths = EXT:tea/Resources/Private/Partials/
-    # cat=plugin.tx_tea/file; type=string; label=Path to template layouts (FE)
-    layoutRootPaths = EXT:tea/Resources/Private/Layouts/
-  }
-
   persistence {
     # cat=plugin.tx_tea//a; type=string; label=Default storage PID
     storagePid =
--- a/Configuration/TypoScript/constants.typoscript
+++ b/Configuration/TypoScript/constants.typoscript
@@ -1,13 +1,4 @@
 plugin.tx_tea {
-  view {
-    # cat=plugin.tx_tea/file; type=string; label=Path to template root (FE)
-    templateRootPath = EXT:tea/Resources/Private/Templates/
-    # cat=plugin.tx_tea/file; type=string; label=Path to template partials (FE)
-    partialRootPath = EXT:tea/Resources/Private/Partials/
-    # cat=plugin.tx_tea/file; type=string; label=Path to template layouts (FE)
-    layoutRootPath = EXT:tea/Resources/Private/Layouts/
-  }
-
   persistence {
     # cat=plugin.tx_tea//a; type=string; label=Default storage PID
     storagePid =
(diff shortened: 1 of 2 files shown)


Configuration/TypoScript/constants.typoscript
```diff
@@ -1,13 +1,4 @@
 plugin.tx_tea {
-  view {
```
linawolf, line 2: To make the templates easily overwritable I would suggest to keep the constants section

Configuration/TypoScript/setup.typoscript
```diff
@@ -1,8 +1,16 @@
 plugin.tx_tea {
   view {
-    templateRootPath = {$plugin.tx_tea.view.templateRootPath}
-    partialRootPath = {$plugin.tx_tea.view.partialRootPath}
-    layoutRootPath = {$plugin.tx_tea.view.layoutRootPath}
+    templateRootPaths {
+      0 = EXT:tea/Resources/Private/Templates/
```
linawolf, line 4:
```suggestion
      0 = EXT:tea/Resources/Private/Templates/
      10 = {$plugin.tx_tea.view.templateRootPath}
```
And use the templates from the constant to override the ones provided by the extension

Configuration/TypoScript/constants.typoscript
oliverklee: They are already overwritable (if I read [the docs](https://docs.typo3.org/m/typo3/book-extbasefluid/10.4/en-us/b-ExtbaseReference/Index.html#view) correctly): If a TypoScript entry with a higher key is present, it will take precedence.

linawolf: yes they are. However It is considered best practice by most extensions to provide a constant that makes overriding easier then fiddling with keys. Also this gives the advantage that the Constant Editor can be used to set those. The constants come with a description which is also like in-vito documentation. News is doing it like that: https://github.com/georgringer/news/blob/main/Configuration/TypoScript/constants.typoscript#L25, Powermail is doing it like that: https://github.com/einpraegsam/powermail/blob/develop/Configuration/TypoScript/Main/constants.typoscript#L5, SF Event Manager also: https://github.com/derhansen/sf_event_mgt/blob/main/Configuration/TypoScript/constants.typoscript#L14

Also when I am writing an extension extending tea but want to make sure that the sitepackage can still override my extensions templates I would choose a key that lies between 0 (your base template) and 10 (the sitepackage) and therefore do not have to depend on keys that someone writing a sitepackage chose.

linawolf:
Also I would not consider this change a bufix if we where in a real production Extension this change would change the way templates can be used in sitepackages. Both are correct, compare [partialRootPath](https://docs.typo3.org/m/typo3/reference-typoscript/main/en-us/ContentObjects/Fluidtemplate/Index.html#id16) and https://docs.typo3.org/m/typo3/reference-typoscript/main/en-us/ContentObjects/Fluidtemplate/Index.html#partialrootpaths

DanielSiepmann:
The single form should have been dropped in 7.0 already: https://docs.typo3.org/c/typo3/cms-core/main/en-us/Changelog/7.0/Breaking-62673-ExtbaseDeprecatedCodeRemoved.html
 
> ActionController
> 
> Support for old view configuration options templateRootPath, layoutRootPath and partialRootPath is dropped. Use the new options with fallback mechanism.

oliverklee:
This is currently blocked by the functional tests failures in TYPO3 12.

oliverklee (approved):

Issue #553, opened 2022-09-04
Make records hidable and timeable


oliverklee:
… with `hidden` and `startdate`/`enddate`.

Spun off from #549.

linawolf:
@oliverklee shall we add 'fe_users' as it is another of the classic enable fields?

oliverklee:
@linawolf Let's have that in a separate ticket. Would you mind creating one?

PR #735, opened 2023-01-24
[TASK] Make records hidable and timeable

lukaszuznanski:
Resolves: #553


lukaszuznanski:
Different TCA needed for v11 and for v12

oliverklee (changes requested):
I've manually tested this with my 11LTS testing installation:

- Hiding works if I use the "hide" icon in the list.
- When I edit a tea record in the DB, the fields hidden, starttime and endtime are not visible yet in the TCEforms.

So we'll also need to add those three fields to the `types` configuration.

oliverklee (changes requested):
With this change, I get this exception when I try to edit a tea record in the BE:

> Item 0 of field hidden of TCA table tx_tea_domain_model_product_tea has no label

lukaszuznanski:
@oliverklee ok, so it's different for v12 and for v11, I have to adjust configuration. It works fine on v12.

Configuration/TCA/tx_tea_domain_model_product_tea.php
```diff
@@ -10,11 +10,53 @@
...
+            'config' => [
+                'type' => 'check',
+                'renderType' => 'checkboxToggle',
+                'items' => [
+                    [
+                        'label' => '',
+                        'invertStateDisplay' => true,
+                    ],
+                ],
+            ],
```
oliverklee, line 35: This now causes the toggle checkbox to work in inverse. With this setting, the record is visible for me:

grafik

oliverklee: This is what the TCA for core records looks like in V11:

```
        'hidden' => [
            'exclude' => true,
            'label' => 'LLL:EXT:core/Resources/Private/Language/locallang_general.xlf:LGL.enabled',
            'config' => [
                'type' => 'check',
                'renderType' => 'checkboxToggle',
                'items' => [
                    [
                        0 => '',
                        'invertStateDisplay' => true,
                    ],
                ],
            ],
        ],
```

oliverklee: Strange, this is the same as our code. I don't understand this.

oliverklee (changes requested):

oliverklee:
Testing in 11LTS, the "hide" toggle still is the wrong way around.

Configuration/TCA/tx_tea_domain_model_product_tea.php
```diff
@@ -10,11 +10,53 @@
...
+        'hidden' => [
+            'exclude' => true,
+            'label' => 'LLL:EXT:core/Resources/Private/Language/locallang_general.xlf:LGL.enabled',
+            'config' => [
+                'type' => 'check',
+                'renderType' => 'checkboxToggle',
+                'items' => [
+                    [
+                        'label' => '',
+                        'invertStateDisplay' => true,
```
oliverklee, line 32: Ah, we also need this added after line 136:

```
'invertStateDisplay' => true,
```

oliverklee (changes requested):

oliverklee:
This still needs the V11 fix, and also a reabse.

oliverklee:
@lukaszuznanski Do you have time to get this PR finished, or would you like me to take over?

lukaszuznanski:
@oliverklee if you have time to finish it, then go for it

oliverklee:
This looks really hard to rebase (also because we've merged `main` into this branch instead of rebasing it). I'm willing to do the V11 fixed, but don't have the time to first untangle the rebase mess right now.

lukaszuznanski:
I think it might be easier and faster to open new PR with those changes, will do.

oliverklee:
> I think it might be easier and faster to open new PR with those changes, will do.

Thanks!

lukaszuznanski:
Closed as there is new PR #989

Issue #551, opened 2022-09-04
Make the records localizable


oliverklee:
Spun off from #549.

PR #751, opened 2023-02-21
Make the records localizable

lukaszuznanski:
Resolves: #551

### DIFF
--- a/Configuration/TCA/tx_tea_domain_model_product_tea.php
+++ b/Configuration/TCA/tx_tea_domain_model_product_tea.php
@@ -48,7 +48,8 @@ $tca = [
                 'renderType' => 'selectSingle',
                 'items' => [
                     [
-                        '', 0,
+                        'label' => '',
+                        'value' => 0,
                     ],
                 ],
                 'foreign_table' => 'tx_tea_domain_model_product_tea',
@@ -152,6 +153,13 @@ if ($typo3Version->getMajorVersion() < 12) {
         ]
     );
     unset($tca['columns']['title']['required']);
+
+    $tca['columns']['l18n_parent']['config']['items'] = [
+        [
+            0 => '',
+            1 => 0,
+        ],
+    ];
     $tca['columns']['image'] = [
         'label' => 'LLL:EXT:tea/Resources/Private/Language/locallang_db.xlf:tx_tea_domain_model_product_tea.image',
         'config' => \TYPO3\CMS\Core\Utility\ExtensionManagementUtility::getFileFieldTCAConfig(
--- a/Configuration/TCA/tx_tea_domain_model_product_tea.php
+++ b/Configuration/TCA/tx_tea_domain_model_product_tea.php
@@ -48,14 +48,13 @@ $tca = [
                 'renderType' => 'selectSingle',
                 'items' => [
                     [
-                        '',
-                        0,
+                        '', 0,
                     ],
                 ],
-                'foreign_table' => 'tx_myextension_domain_model_something',
+                'foreign_table' => 'tx_tea_domain_model_product_tea',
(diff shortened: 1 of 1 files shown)


Configuration/TCA/tx_tea_domain_model_product_tea.php
```diff
@@ -10,11 +10,52 @@
...
+            'displayCond' => 'FIELD:sys_language_uid:>:0',
+            'label' => 'LLL:EXT:core/Resources/Private/Language/locallang_general.xlf:LGL.l18n_parent',
+            'config' => [
+                'type' => 'select',
+                'renderType' => 'selectSingle',
+                'items' => [
+                    [
+                        '',
+                        0,
+                    ],
```
oliverklee, line 39: Please let's have this in one line:
```
['', 0],
```

Configuration/TCA/tx_tea_domain_model_product_tea.php
```diff
@@ -10,11 +10,52 @@
...
+                ],
+                'foreign_table' => 'tx_myextension_domain_model_something',
+                'foreign_table_where' =>
+                    'AND {#tx_myextension_domain_model_something}.{#pid}=###CURRENT_PID###'
+                    . ' AND {#tx_myextension_domain_model_something}.{#sys_language_uid} IN (-1,0)',
```
oliverklee, line 44: `#tx_myextension_domain_model_something` looks like a copy'n'paste error from a template.

oliverklee (changes requested):

oliverklee:
I currently do not have a multi-language testing installation in which I could test this PR.

@brotkrueml, @DanielSiepmann Do you have a multi-language installation where you could test and review this PR?

DanielSiepmann:
> I currently do not have a multi-language testing installation in which I could test this PR.
> 
> @brotkrueml, @DanielSiepmann Do you have a multi-language installation where you could test and review this PR?

Why not add functional tests (using InternalRequest) that ensure the rendering will work? This can easily be used to make the testing instance available to the browser and manually test the backend.
Can't promise anything, but will try to add this next week if you like.

We still don't demonstrate the InternalRequest within Functional Tests anyway. So would be an additional bonus?

lukaszuznanski:
Ok so I should add tests to this PR so we can proceed.

DanielSiepmann:
I wouldn't say should. That would also only verify frontend rendering based on DB content. It wouldn't check the editing experience within Backend. Maybe we should move that to a separate issue to keep this to a minimum. I'll test your change in v11 and 12 now.

DanielSiepmann:
@lukaszuznanski I've rebased on current main and solves the TCA deprecation.
I've also tested BE and FE with v12 and v11. Looks good to me.

We don't show the language field in records, but that's not necessary anyway.

oliverklee (changes requested):
Looks and works generally good.

Still, I think we should only make the text fields (title and description) available in the translated records, and use the image from the original record. What do you think?

DanielSiepmann:
> Still, I think we should only make the text fields (title and description) available in the translated records, and use the image from the original record. What do you think?

TYPO3 allows to have a radio to choose whether to use the one from default or a custom one. That would for example allow to use the default one in case you have a picture of the tea itself. But one could use a translated version in case the picture is the packaging with text on it and available in multiple languages?

That way we also can demonstrate that TYPO3 feature.

oliverklee (approved):

oliverklee:
> TYPO3 allows to have a radio to choose whether to use the one from default or a custom one. That would for example allow to use the default one in case you have a picture of the tea itself. But one could use a translated version in case the picture is the packaging with text on it and available in multiple languages?

Then let's have that as a separate PR.

PR #787, opened 2023-04-01
[FEATURE] Support PHP 8.2

brotkrueml:


oliverklee (changes requested):
We also should add PHP 8.2 to the GitLab CI configuration.

oliverklee (approved):

PR #798, opened 2023-04-06
[FEATURE] Allow to hide teas based on fe_groups

### LINKED ISSUE #555: Allow making records  visible only for certain fe_groups
With 
$GLOBALS['TCA'][$table]['ctrl']['enablecolumns']['fe_group']

https://docs.typo3.org/m/typo3/reference-tca/main/en-us/Ctrl/Properties/Enablecolumns.html

DanielSiepmann:
Resolves #555

### DIFF
--- a/Configuration/TCA/tx_tea_domain_model_product_tea.php
+++ b/Configuration/TCA/tx_tea_domain_model_product_tea.php
@@ -10,9 +10,23 @@ $tca = [
         'default_sortby' => 'title',
         'iconfile' => 'EXT:tea/Resources/Public/Icons/Record.svg',
         'searchFields' => 'title, description',
+        'enablecolumns' => [
+            'fe_group' => 'fe_group',
+        ],
     ],
     'types' => [
-        '1' => ['showitem' => 'title, description, image'],
+        '1' => ['showitem' => '
+            --div--;LLL:EXT:core/Resources/Private/Language/Form/locallang_tabs.xlf:general,
+                title, description, image,
+            --div--;LLL:EXT:tea/Resources/Private/Language/locallang_db.xlf:tx_tea_domain_model_product_tea.tabs.access,
+                --palette--;;access,
+        '],
+    ],
+    'palettes' => [
+        'access' => [
+            'label' => 'LLL:EXT:tea/Resources/Private/Language/locallang_db.xlf:tx_tea_domain_model_product_tea.palettes.access',
+            'showitem' => 'fe_group',
+        ],
     ],
     'columns' => [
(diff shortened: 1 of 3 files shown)


DanielSiepmann:
Tested with v11.5 and v12.

V12 allows you to add array keys, but that wouldn't work in v11, so I didn't add them.

brotkrueml:
> Tested with v11.5 and v12.
> 
> V12 allows you to add array keys, but that wouldn't work in v11, so I didn't add them.
> 
> That's what is reported within the tests. Should I add a condition to provide different config based on version?
> 
> `The TCA field 'fe_group' of table 'tx_tea_domain_model_product_tea' uses the legacy way of defining 'items'. Please switch to associated array keys: label, value, icon, group, description.`

You can add both: 0, 1, label, value. Then the deprecation should be avoided.

DanielSiepmann:
> You can add both: 0, 1, label, value. Then the deprecation should be avoided.

Thanks @brotkrueml for your suggestion, but it doesn't help.

brotkrueml:
> > You can add both: 0, 1, label, value. Then the deprecation should be avoided.
> 
> Thanks @brotkrueml for your suggestion, but it doesn't help.

Sorry, then you might need something like this to avoid the deprecation:

https://github.com/brotkrueml/typo3-matomo-widgets/blob/main/Configuration/SiteConfiguration/Overrides/sites.php#L79-L86

I use label/value by default and then change to 0/1 when v11.

DanielSiepmann:
The phpunit issue with 12.3 seems unrelated to my change. I don't know what's causing them.

oliverklee:
The PHPUnit failure is in deed unrelated, and we need to solve it first. The cause is that in V12 `ActionController` does not have the `forward` or `redirect` methods anymore, and hence we cannot mock them anymore. We'll need to add a version switch for this.

DanielSiepmann:
I can do that @oliverklee, give me some minutes. Still wondering why I don't have the issue locally. But never mind.

oliverklee:
There was a very recent change in the TYPO3 TF switching to `onlyMethods` that made this problem visible. With a composer update, possibly the problem will occur locally as well.

DanielSiepmann:
Looks like composer install without `composer.lock` etc. still doesn't fetch new commits into composer cache. At least that would explain why removing all files and running `composer install` didn't reveal the issue. I now ran `composer update` and could reproduce, fix: https://github.com/TYPO3-Documentation/tea/pull/799.

oliverklee (approved):

Issue #721, opened 2023-01-11
Provide a contribution guide


DanielSiepmann:
Not everyone might now how we work and how GitHub works.
We therefore should provide a contribution guide. GitHub provides a link to a guide if we follow the docs: https://docs.github.com/en/communities/setting-up-your-project-for-healthy-contributions/setting-guidelines-for-repository-contributors

The guide should probably explain how to contribute in general. https://docs.typo3.org/Home/Contribute.html does already exist, but contribution to docs and TYPO3 core is different from contributing to TYPO3 extensions on GitHub and GitLab. This workflow should be covered. E.g. First create a fork, followed by a new branch. Apply your changes and create a pull request.

This should lower the barrier for others to contribute. Especially those who are not using GitHub on a daily basis.

DanielSiepmann:
Done with https://github.com/FriendsOfTYPO3/tea/pull/1044, available at: https://github.com/FriendsOfTYPO3/tea/blob/main/CONTRIBUTING.md

PR #810, opened 2023-04-12
[TASK] Add contribution guide

DanielSiepmann:
Resolves: #721


DanielSiepmann:
@sbuerk would be cool if you could review the change, as it is based on your feedback from December.

sbuerk (approved):
This is at least good first start. This can be enriched later - or linking to repo internal enhanced documentation about some topics for the contribution.

Only one thing, I'm not really sure about, is the filename.

As this is markdown, it may be use-full to use CONTRIBUTING.md. TYPO3 core has that naming too.

https://github.com/TYPO3/typo3/blob/main/CONTRIBUTING.md

Approving for now, but please reconsider the file naming with your team.

DanielSiepmann:
> As this is markdown, it may be use-full to use CONTRIBUTING.md. TYPO3 core has that naming too.

Thanks will adjust, https://docs.github.com/en/communities/setting-up-your-project-for-healthy-contributions/setting-guidelines-for-repository-contributors is misleading as it doesn't have a file extension. But linked examples also end with `.md`.

CONTRIBUTING.md
```diff
@@ -0,0 +1,20 @@
...
+1. Please check existing open and closed issues first and feel free to comment and re-open them.
+2. Use a meaningful title
+3. Provide enough context, a concrete use case and steps to reproduce bugs
+
+## Contributing code and documentation
+
+1. Create a fork of the repository
+2. Create a new branch holding your changes
+3. Apply your changes
+4. Commit your changes, follow the commit conventions: https://docs.typo3.org/m/typo3/guide-contributionworkflow/main/en-us/Appendix/CommitMessage.html#commitmessage
```
lukaszuznanski, line 16: this guide is directly for gerrit, should we mention it? Not all of those conventions are relevant for github

DanielSiepmann: Fair point, will adjust. I set PR to draft until I've finished.

PR #823, opened 2023-04-25
[TASK] Switch to a stable versions of `typo3/testing-framework`

oliverklee:
Now that version 7.0.0 and 8.0.0 have been released, we can switch away from the dev version.

The new version also makes the testing framework bridge unnecessary.


composer.json
```diff
@@ -46,13 +46,12 @@
...
 		"phpunit/phpunit": "^9.6.7",
 		"saschaegerer/phpstan-typo3": "^1.8.4",
-		"sbuerk/typo3-cmscomposerinstallers-testingframework-bridge": "^0.1.3",
 		"seld/jsonlint": "^1.9.0",
 		"squizlabs/php_codesniffer": "^3.7.2",
 		"symfony/yaml": "^5.4 || ^6.1",
 		"typo3/cms-fluid-styled-content": "^11.5.4 || ^12.4",
 		"typo3/coding-standards": "^0.6.1",
-		"typo3/testing-framework": "^7.0@dev"
+		"typo3/testing-framework": "^7.0.0"
```
brotkrueml, line 54: Out of curiosity: Why `^7.0.0` and not just `^7.0` (like for the other the requirements)?

oliverklee: I'd like to also control the bug-fix release version of the dev dependencies. (And I think I'd like to change the other dev dependencies accordingly except for `symfony/yaml`, which also is an (indirect) production dependency and which we'd hence also like to check with the lowest/highest dependency versions.)

brotkrueml: Okay, but `^7.0` and `^7.0.0` are equivalent. But fine for me.

brotkrueml (approved):

Issue #826, opened 2023-05-01
Adjust ext_localconf.php / ext_tables.php


linawolf:
After dropping TYPO3 10.4 support it is considered best practice to  make us of the automatically wrapping into a global namespace of the configuration files. 

- Using an anonymous function is not needed anymore
- `declare(strict_types=1);` can and should be used
- Use statements instead of FQN should be used

PR #833, opened 2023-05-08
[TASK] Cleanup ext_localconf.php

linawolf:
After dropping TYPO3 10.4 support:

- Using an anonymous function is not needed anymore
- declare(strict_types=1); can and should be used
- Use statements instead of FQN should be used

resolves https://github.com/TYPO3-Documentation/tea/issues/826

### DIFF
--- a/ext_localconf.php
+++ b/ext_localconf.php
@@ -1,31 +1,34 @@
 <?php
 
+declare(strict_types=1);
+
+use TTN\Tea\Controller\TeaController;
+use TYPO3\CMS\Extbase\Utility\ExtensionUtility;
+
 defined('TYPO3') or die('Access denied.');
 
-(static function (): void {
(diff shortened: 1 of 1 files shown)


ext_localconf.php
```diff
@@ -1,31 +1,34 @@
...
-            \TTN\Tea\Controller\TeaController::class => 'show',
-        ],
-        [
-            \TTN\Tea\Controller\TeaController::class => '',
-        ]
-    );
-})();
+// This makes the plugin available for front-end rendering.
+ExtensionUtility::configurePlugin(
+// extension name, matching the PHP namespaces (but without the vendor)
```
brotkrueml, line 12:
```suggestion
    // extension name, matching the PHP namespaces (but without the vendor)
```

brotkrueml (approved):

PR #856, opened 2023-06-19
[TASK] Make the controller tests more fine-grained

oliverklee:
Tests should only test one thing (if possible).


DanielSiepmann (approved):

DanielSiepmann:
@oliverklee would you like to have FunctionalTests that execute a request and check the rendered HTML?
Like I've added here: https://github.com/werkraum-media/watchlist/blob/main/Tests/Functional/FormIntegrationTest.php#L58
I then would open an issue and propose a PR.

oliverklee:
> would you like to have FunctionalTests that execute a request and check the rendered HTML?
(…) I then would open an issue and propose a PR.

Yes, please, that would be awesome. Thanks! :heart:

DanielSiepmann:
Issue created: https://github.com/TYPO3-Documentation/tea/issues/859

PR #860, opened 2023-06-20
[FEATURE] Add `Tea.ownerUid`

oliverklee:
This is a pre-patch for adding a CRUD plugin for tea records.

This property contains only the UID of the owner FE user, but not a relation to a FE user model. This is because we neither have nor need a FE user model for the purposes of the CRUD plugin.

### DIFF
--- a/Classes/Domain/Model/Product/Tea.php
+++ b/Classes/Domain/Model/Product/Tea.php
@@ -25,6 +25,13 @@ class Tea extends AbstractEntity
      */
     protected $image;
 
+    // Note: We cannot use `@var` for the more specific type annotation here as this confuses the Extbase type mapper.
+
+    /**
+     * @phpstan-var int<0, max>
+     */
+    protected int $ownerUid = 0;
+
     public function getTitle(): string
     {
         return $this->title;
@@ -59,5 +66,21 @@ class Tea extends AbstractEntity
     public function setImage(FileReference $image): void
     {
         $this->image = $image;
+    }
+
+    /**
+     * @return int<0, max>
+     */
+    public function getOwnerUid(): int
+    {
+        return $this->ownerUid;
+    }
+
+    /**
+     * @param int<0, max> $ownerUid
+     */
+    public function setOwnerUid(int $ownerUid): void
+    {
+        $this->ownerUid = $ownerUid;
     }
 }
--- a/Configuration/Extbase/Persistence/Classes.php
+++ b/Configuration/Extbase/Persistence/Classes.php
@@ -1,0 +1,11 @@
+<?php
+
+declare(strict_types=1);
+
+return [
+    \TTN\Tea\Domain\Model\Product\Tea::class => [
+        'properties' => [
+            'ownerUid' => ['fieldName' => 'owner'],
+        ],
+    ],
+];
--- a/Configuration/TCA/tx_tea_domain_model_product_tea.php
+++ b/Configuration/TCA/tx_tea_domain_model_product_tea.php
@@ -21,7 +21,7 @@ $tca = [
     'types' => [
         '1' => ['showitem' => '
             --div--;LLL:EXT:core/Resources/Private/Language/Form/locallang_tabs.xlf:general,
-                title, description, image,
+                title, description, image, owner,
             --div--;LLL:EXT:tea/Resources/Private/Language/locallang_db.xlf:tx_tea_domain_model_product_tea.tabs.access,
                 --palette--;;access,
         '],
@@ -133,8 +133,23 @@ $tca = [
                 'foreign_table' => 'fe_groups',
             ],
         ],
+        'owner' => [
+            'exclude' => true,
+            'l10n_mode' => 'exclude',
+            'label' => 'LLL:EXT:tea/Resources/Private/Language/locallang_db.xlf:tx_tea_domain_model_product_tea.owner',
+            'config' => [
+                'type' => 'group',
+                'allowed' => 'fe_users',
+                'default' => 0,
+                'size' => 1,
+                'minitems' => 0,
+                'maxitems' => 1,
+                'hideSuggest' => true,
+            ],
+        ],
     ],
 ];
+
 $typo3Version = new \TYPO3\CMS\Core\Information\Typo3Version();
 if ($typo3Version->getMajorVersion() < 12) {
     $tca = array_replace_recursive(
--- a/Resources/Private/Language/de.locallang_db.xlf
+++ b/Resources/Private/Language/de.locallang_db.xlf
@@ -23,6 +23,10 @@
 				<source>Image</source>
 				<target>Bild</target>
 			</trans-unit>
+			<trans-unit id="tx_tea_domain_model_product_tea.owner">
+				<source>Website user who created this record</source>
+				<target>Website-Benutzer, der diesen Datensatz erstellt hat</target>
+			</trans-unit>
 			<trans-unit id="tx_tea_domain_model_product_tea.fe_group">
 				<source>Usergroup Access Rights</source>
 				<target>Zugriffsrechte für Benutzergruppen</target>
--- a/Tests/Functional/Domain/Repository/Fixtures/Product/Tea.csv
+++ b/Tests/Functional/Domain/Repository/Fixtures/Product/Tea.csv
@@ -1,8 +1,8 @@
 "tx_tea_domain_model_product_tea"
-,"uid","pid","title","description","image"
-,"1","1","Earl Grey","Fresh and hot.",0
-,"2","1","Assam","Dark and strong.",0
-,"3","1","Gunpowder","Bitter and very green.",1
+,"uid","pid","title","description","image","owner"
+,"1","1","Earl Grey","Fresh and hot.",0,2
+,"2","1","Assam","Dark and strong.",0,0
+,"3","1","Gunpowder","Bitter and very green.",1,0
 "sys_file_reference"
 ,"uid","pid","uid_foreign","tablenames","fieldname"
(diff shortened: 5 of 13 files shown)


Classes/Domain/Model/Product/Tea.php
```diff
@@ -60,4 +65,20 @@ public function setImage(FileReference $image): void
     {
         $this->image = $image;
     }
+
+    /**
+     * @return int<0, max>
```
DanielSiepmann, line 70: Do we need the annotations? I'm asking because other methods like `getTitle()` don't have them any more and I always prefer to have as less maintenance as possible.

Configuration/Extbase/Persistence/Classes.php
```diff
@@ -0,0 +1,11 @@
+<?php
+
+declare(strict_types=1);
+
+return [
+    \TTN\Tea\Domain\Model\Product\Tea::class => [
+        'properties' => [
+            'ownerUid' => ['fieldName' => 'owner'],
```
DanielSiepmann, line 8: Can you explain why you don't use the convention but use `ownerUid` and `owner`?

My thought was that the extension some day might switch to a full Owner Model and doesn't need to migrate the database column name in that case. And it is total fine within the database to use `owner` for both as it always will be the uid. But the PHP code should document its own type. Did you have other thoughts?

Configuration/TCA/tx_tea_domain_model_product_tea.php
```diff
@@ -19,12 +19,14 @@
...
+        '1' => [
+            'showitem' => '
             --div--;LLL:EXT:core/Resources/Private/Language/Form/locallang_tabs.xlf:general,
-                title, description, image,
+                title, description, image, owner,
             --div--;LLL:EXT:tea/Resources/Private/Language/locallang_db.xlf:tx_tea_domain_model_product_tea.tabs.access,
                 --palette--;;access,
-        '],
+        ',
+        ],
```
DanielSiepmann, line 29: That indentation looks strange to me. The indentation of closing `'` doesn't seem related to the line containing the opening one. It is on the same level as the outer `[]`.

Resources/Private/Language/de.locallang_db.xlf
```diff
@@ -23,6 +23,10 @@
 				<source>Image</source>
 				<target>Bild</target>
 			</trans-unit>
+			<trans-unit id="tx_tea_domain_model_product_tea.owner">
+				<source>FE user who created this record</source>
+				<target>FE-User, der diesen Datensatz erstellt hat</target>
```
DanielSiepmann, line 28: Maybe we should use the TYPO3 naming for the record type "Website-Benutzer" in German and "Website User" in English instead of "FE-User" and "FE user"? That would provide a streamlined UI/UX for editors.

Tests/Functional/Domain/Repository/Fixtures/Product/Tea.csv
```diff
@@ -1,8 +1,8 @@
 "tx_tea_domain_model_product_tea"
-,"uid","pid","title","description","image"
-,"1","1","Earl Grey","Fresh and hot.",0
-,"2","1","Assam","Dark and strong.",0
-,"3","1","Gunpowder","Bitter and very green.",1
+,"uid","pid","title","description","image","owner"
```
DanielSiepmann, line 2: Nothing for this PR, but a general question: Should we discuss the CSV Format? I know this is the only official way provided by the testing framework. But our company has no good relation to CSV, we prefer PHP.

Maybe something to discuss within our regular team call, what do you think?

Tests/Functional/Domain/Repository/Product/TeaRepositoryTest.php
```diff
@@ -133,4 +134,53 @@ public function addAndPersistAllCreatesNewRecord(): void
         self::assertIsArray($databaseRow);
         self::assertSame($title, $databaseRow['title']);
     }
+
+    /**
+     * @test
+     */
+    public function findByOwnerUidFindsEventWithTheGivenOwnerUid(): void
```
DanielSiepmann, line 141: I guess "Event" is a copy and paste or autocompletion error and should be "Tea" instead?

Tests/Functional/Domain/Repository/Product/TeaRepositoryTest.php
```diff
@@ -133,4 +134,53 @@ public function addAndPersistAllCreatesNewRecord(): void
...
+
+        $result = $this->subject->findByOwnerUid(1);
+
+        self::assertCount(1, $result);
+    }
+
+    /**
+     * @test
+     */
+    public function findByOwnerUidFindsIgnoresEventWithOtherOwnerUid(): void
```
DanielSiepmann, line 153: I guess "Other" is too generic. Maybe something like "findByOwnerUidFindsNothingForNoneExistingOwnerUid"? or "findByOwnerUidFindsNothingForNoneMatchingOwnerUid"

Classes/Domain/Model/Product/Tea.php
oliverklee: In general, we don't have the PHPDoc type annotations if they do not add anything over the native type declaration (neither more precise type information nor a relevant human-readable description). In this case, we make the type more precise, which in turn helps PHPStan detect incorrect calls: We change it from the native type declaration `int` to the more specific `int<0, max>` (i.e., a non-negative int).

I'm using `@phpstan-var` instead of `@var` for the property as having this type annotation confused Extbase.

(I do not insist on having these more precise type annotations. I'd prefer them, but if they block the +1, I can remove them.)

Configuration/Extbase/Persistence/Classes.php
oliverklee: By convention, we never have the `*_uid` suffix for associations in the DB column names. This is why I'm choosing `owner` instead of `owner_uid`.

On the PHP side, things are different. An `owner` implicates that this is an object (in this case, a FE user model), not an int (a UID). To make it clear that we're dealing with the UID and not the model itself (and hence reduce the amount of surprises in the code), I've chosen the name `ownerUid` instead of `owner`. (Also, this indeed leaves the way open for us to switch to an association to a model using the association name `owner` later if the need arises.)

Configuration/TCA/tx_tea_domain_model_product_tea.php
oliverklee: I'll undo this formatting change and move it (in a cleaned-up version) to a separate PR. I think this was some fallout from me having autoformatted the file.

Resources/Private/Language/de.locallang_db.xlf
oliverklee: Thanks, done.

Tests/Functional/Domain/Repository/Fixtures/Product/Tea.csv
oliverklee: Yes, let's move this discussion to our call.

Tests/Functional/Domain/Repository/Product/TeaRepositoryTest.php
oliverklee: Done.

Classes/Domain/Model/Product/Tea.php
DanielSiepmann: Thanks for the info. Maybe that's something we should document somewhere if not already done. I remember there was a discussion about that regarding TYPO3 core source code as well? Are you involved in that topic? Maybe that's nothing to document for EXT:tea but https://docs.typo3.org/m/typo3/reference-coreapi/12.4/en-us/CodingGuidelines/CglPhp/CodingBestPractices/Index.html?

Configuration/Extbase/Persistence/Classes.php
DanielSiepmann: Thanks for clarification, that resolves the conversation for me.

Tests/Functional/Domain/Repository/Fixtures/Product/Tea.csv
DanielSiepmann: Done: https://notes.typo3.org/JKT4Cr6tSZehjq0Nu8o65g?both#discussion-keep-csv-for-database-fixtures

Classes/Domain/Model/Product/Tea.php
oliverklee: Yes, we should document this there.

I'm also going to add a comment here in our code explaining why I'm using `@phpstan-var` here.

oliverklee: Done.

Tests/Functional/Domain/Repository/Product/TeaRepositoryTest.php
DanielSiepmann: I still see "Event" at https://github.com/TYPO3-Documentation/tea/pull/860/files#diff-516bb26cd1c2b4fe34e6ef45b6a15818f42846677642456d762045c06419fcc0R153, did I miss something? (I don't do reviews on GitHub that often) or did you miss it while pushing?

oliverklee: Argh, sorry, my mistake. There were indeed multiple occurrences (probably from copy'n'pasting), and I only fixed one. Done and repushed.

DanielSiepmann (approved):

Tests/Functional/Domain/Repository/Product/TeaRepositoryTest.php
DanielSiepmann: Okay, I'll try to remember and mention it next time :)

PR #863, opened 2023-06-21
[TASK] Restructure the label keys

oliverklee:
Label keys should be specific to their context. This allows having different labels for the same thing in different contexts (e.g., having different headings for the tea title in the regular list view and the FE editor).

We take the plugin as the context here, which allows us to reuse labels between different actions of the same plugin (e.g., between the list view and single view).

### DIFF
--- a/Resources/Private/Language/de.locallang.xlf
+++ b/Resources/Private/Language/de.locallang.xlf
@@ -11,15 +11,15 @@
 				<source>Tea single view</source>
 				<target>Tee-Einzelansicht</target>
 			</trans-unit>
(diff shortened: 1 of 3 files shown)


DanielSiepmann (approved):

DanielSiepmann:
Approved. We already have different files per context, e.g. `locallang_db.xlf` for TCA/Database. But I guess this is a preparation for further actions (CRUD).

oliverklee:
> But I guess this is a preparation for further actions (CRUD).

It is indeed!

PR #864, opened 2023-06-21
[FEATURE] Add a list view for the FE tea editor

oliverklee:
This is one more part of building CRUD functionality.

### DIFF
--- a/Classes/Controller/FrontEndEditorController.php
+++ b/Classes/Controller/FrontEndEditorController.php
@@ -1,0 +1,44 @@
+<?php
+
+declare(strict_types=1);
+
+namespace TTN\Tea\Controller;
+
+use Psr\Http\Message\ResponseInterface;
+use TTN\Tea\Domain\Repository\Product\TeaRepository;
+use TYPO3\CMS\Core\Context\Context;
+use TYPO3\CMS\Extbase\Mvc\Controller\ActionController;
+
+/**
+ * Controller for a CRUD FE editor for teas.
+ */
+class FrontEndEditorController extends ActionController
+{
+    private Context $context;
+
+    private TeaRepository $teaRepository;
+
+    public function __construct(Context $context, TeaRepository $teaRepository)
+    {
+        $this->context = $context;
+        $this->teaRepository = $teaRepository;
+    }
+
+    public function indexAction(): ResponseInterface
+    {
+        $userUid = $this->getUidOfLoggedInUser();
+        if ($userUid > 0) {
+            $this->view->assign('teas', $this->teaRepository->findByOwnerUid($userUid));
+        }
+
+        return $this->htmlResponse();
+    }
+
+    /**
+     * @return int<0, max>
+     */
+    private function getUidOfLoggedInUser(): int
+    {
+        return $this->context->getPropertyFromAspect('frontend.user', 'id');
+    }
+}
--- a/Resources/Private/Layouts/FrontEndEditor.html
+++ b/Resources/Private/Layouts/FrontEndEditor.html
@@ -1,0 +1,12 @@
+<html xmlns:f="http://typo3.org/ns/TYPO3/CMS/Fluid/ViewHelpers" data-namespace-typo3-fluid="true">
+    <f:security.ifAuthenticated>
+        <f:then>
+            <f:render section="main"/>
+        </f:then>
+        <f:else>
+            <div class="alert alert-danger" role="alert">
+                Please configure this plugin to be only visible if a website user is logged in.
+            </div>
+        </f:else>
+    </f:security.ifAuthenticated>
+</html>
--- a/Tests/Unit/Controller/FrontEndEditorControllerTest.php
+++ b/Tests/Unit/Controller/FrontEndEditorControllerTest.php
@@ -1,0 +1,121 @@
+<?php
+
+declare(strict_types=1);
+
+namespace TTN\Tea\Tests\Unit\Controller;
+
+use PHPUnit\Framework\MockObject\MockObject;
+use TTN\Tea\Controller\FrontEndEditorController;
+use TTN\Tea\Domain\Repository\Product\TeaRepository;
+use TYPO3\CMS\Core\Context\Context;
+use TYPO3\CMS\Core\Context\UserAspect;
+use TYPO3\CMS\Core\Http\HtmlResponse;
+use TYPO3\CMS\Core\Information\Typo3Version;
+use TYPO3\CMS\Extbase\Mvc\Controller\ActionController;
+use TYPO3\CMS\Extbase\Persistence\QueryResultInterface;
+use TYPO3\CMS\Fluid\View\TemplateView;
+use TYPO3\TestingFramework\Core\AccessibleObjectInterface;
+use TYPO3\TestingFramework\Core\Unit\UnitTestCase;
+
+/**
+ * @covers \TTN\Tea\Controller\FrontEndEditorController
+ */
+final class FrontEndEditorControllerTest extends UnitTestCase
+{
+    /**
+     * @var FrontEndEditorController&MockObject&AccessibleObjectInterface
+     */
+    private FrontEndEditorController $subject;
+
+    /**
+     * @var TemplateView&MockObject
+     */
+    private TemplateView $viewMock;
+
+    private Context $context;
+
+    /**
+     * @var TeaRepository&MockObject
+     */
+    private TeaRepository $teaRepositoryMock;
+
+    protected function setUp(): void
+    {
+        parent::setUp();
+
+        $this->context = new Context();
+        $this->teaRepositoryMock = $this->createMock(TeaRepository::class);
+
+        // We need to create an accessible mock in order to be able to set the protected `view`.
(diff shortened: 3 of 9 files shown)


Classes/Controller/FrontEndEditorController.php
```diff
@@ -0,0 +1,34 @@
...
+ * Controller for a CRUD FE editor for teas.
+ */
+class FrontEndEditorController extends ActionController
+{
+    private TeaRepository $teaRepository;
+
+    public function injectTeaRepository(TeaRepository $teaRepository): void
+    {
+        $this->teaRepository = $teaRepository;
+    }
```
DanielSiepmann, line 23: Is there a reason not use constructor injection?
The `ActionController` has moved everything to method injection in order to free constructor injection for final controller.

Classes/Controller/FrontEndEditorController.php
```diff
@@ -0,0 +1,34 @@
...
+
+    public function indexAction(): ResponseInterface
+    {
+        $userUid = GeneralUtility::makeInstance(Context::class)->getPropertyFromAspect('frontend.user', 'id');
```
DanielSiepmann, line 27: Is there a reason not to inject the Context but to use `GeneralUtility`?

Tests/Unit/Controller/FrontEndEditorControllerTest.php
```diff
@@ -0,0 +1,117 @@
...
+        $methodsToMock = ['htmlResponse', 'redirect', 'redirectToUri'];
+        if ((new Typo3Version())->getMajorVersion() <= 11) {
+            $methodsToMock[] = 'forward';
+        }
+        $this->subject = $this->getAccessibleMock(FrontEndEditorController::class, $methodsToMock);
+
+        $this->viewMock = $this->createMock(TemplateView::class);
+        $this->subject->_set('view', $this->viewMock);
+
+        $this->teaRepositoryMock = $this->getMockBuilder(TeaRepository::class)->disableOriginalConstructor()->getMock();
```
DanielSiepmann, line 59: You are the export regarding unit testing :) So I use this opportunity to ask you what's the benefit of this approach over `->createStub()`?

Tests/Unit/Controller/FrontEndEditorControllerTest.php
```diff
@@ -0,0 +1,117 @@
...
+        $this->subject->injectTeaRepository($this->teaRepositoryMock);
+
+        $responseMock = $this->createMock(HtmlResponse::class);
```
DanielSiepmann, line 62: Shouldn't that be a `createStub()` as there are no assertions on `$responseMock`?

Tests/Unit/Controller/FrontEndEditorControllerTest.php
```diff
@@ -0,0 +1,117 @@
...
+    {
+        self::assertInstanceOf(ActionController::class, $this->subject);
+    }
+
+    /**
+     * @test
+     */
+    public function indexActionForNoLoggedInUserAssignsNothingToView(): void
+    {
+        $userAspect = $this->createMock(UserAspect::class);
```
DanielSiepmann, line 82: Shouldn't that be a `createStub()` as there are no assertions on `$userAspect`?

Tests/Unit/Controller/FrontEndEditorControllerTest.php
```diff
@@ -0,0 +1,117 @@
...
+        $this->subject = $this->getAccessibleMock(FrontEndEditorController::class, $methodsToMock);
+
+        $this->viewMock = $this->createMock(TemplateView::class);
+        $this->subject->_set('view', $this->viewMock);
+
+        $this->teaRepositoryMock = $this->getMockBuilder(TeaRepository::class)->disableOriginalConstructor()->getMock();
+        $this->subject->injectTeaRepository($this->teaRepositoryMock);
+
+        $responseMock = $this->createMock(HtmlResponse::class);
+        $this->subject->method('htmlResponse')->willReturn($responseMock);
```
DanielSiepmann, line 63: Some Variables end with a suffix `Mock` others don't (Compare to `$userAspect`). I'd prefer a rule that's always applied. Either always add the suffix or don't add it at all.

Resources/Private/Layouts/FrontEndEditor.html
```diff
@@ -0,0 +1,13 @@
+<html xmlns:f="http://typo3.org/ns/TYPO3/CMS/Fluid/ViewHelpers" data-namespace-typo3-fluid="true">
+    <f:security.ifAuthenticated>
+        <f:then>
+            <f:render section="main"/>
+        </f:then>
+        <f:else>
+            <div class="alert alert-danger" role="alert">
+                Please configure this plugin to be only visible if a FE user in the group with the
```
DanielSiepmann, line 8: The message says `in the group` but the check is against any authentication.
Will that change with an upcoming PR, should we change the message or the check?

Classes/Controller/FrontEndEditorController.php
oliverklee: Thanks, good point! Done.

Classes/Controller/FrontEndEditorController.php
oliverklee: No - only the reasons that I didn't think of it. ;-) Done.

Tests/Unit/Controller/FrontEndEditorControllerTest.php
oliverklee: `createStub()` indeed is the better choice for our mocks here. (I had completely missed out on this method when it got introduced.) Thanks for pointing this out!

Tests/Unit/Controller/FrontEndEditorControllerTest.php
oliverklee: Indeed. Done.

Tests/Unit/Controller/FrontEndEditorControllerTest.php
oliverklee: Good point. Done.

Resources/Private/Layouts/FrontEndEditor.html
oliverklee: Thanks, I've updated the message now. (It was a copy'n'paste thing.)

oliverklee:
All done and repushed. (I've changed the stubs that actually had expectations back to mocks as PHPStan complained about missing methods. I don't know whether PHPStan is correct or this is something missing from the PHPUnit PHPStan extension, though.)

Tests/Unit/Controller/FrontEndEditorControllerTest.php
```diff
@@ -0,0 +1,115 @@
...
+        $this->subject = $this->getAccessibleMock(
+            FrontEndEditorController::class,
+            $methodsToMock,
+            [$this->context, $this->teaRepositoryMock]
+        );
+
+        $this->viewMock = $this->createMock(TemplateView::class);
+        $this->subject->_set('view', $this->viewMock);
+
+        $responseMock = $this->createStub(HtmlResponse::class);
```
DanielSiepmann, line 63: Should we rename `$responseMock` to `$responseStub`, now that it is a stub?

Tests/Unit/Controller/FrontEndEditorControllerTest.php
DanielSiepmann: This is still `createMock()`, due to using `->expects()` I guess which is not expected on stubs as this is an assertion?
Should we remove the expects here as our real assertion and focus is on view assignment?

DanielSiepmann:
> I don't know whether PHPStan is correct or this is something missing from the PHPUnit PHPStan extension,

I think it is conceptionally correct but technically incorrect. But that way it ensures that one uses the expected create Method.

Tests/Unit/Controller/FrontEndEditorControllerTest.php
oliverklee: Done.

oliverklee:
Repushed.

I've also introduced `FrontEndEditorController::getUidOfLoggedInUser` and `FrontEndEditorControllerTest::setUidOfLoggedInUser`

oliverklee:
> This is still `createMock()`, due to using `->expects()` I guess which is not expected on stubs as this is an assertion?

Exactly.

I have removed the `expects(self::once())` call now. We still need to keep using `createMock` (instead of `createStub`) as PHPStan otherwise won't accept the `with('id')`, which I really want to keep.

Tests/Unit/Controller/FrontEndEditorControllerTest.php
oliverklee: > This is still `createMock()`, due to using `->expects()` I guess which is not expected on stubs as this is an assertion?

Exactly.

I have removed the `expects(self::once())` call now. We still need to keep using `createMock` (instead of `createStub`) as PHPStan otherwise won't accept the `with('id')`, which I really want to keep.

DanielSiepmann: Okay, fine for me.

DanielSiepmann (approved):

PR #866, opened 2023-06-21
[FEATURE] Add validation to the FE-editable tea properties

oliverklee:
The maximum length of the string fields ensures that the contents
fit within the limits of the corresponding DB columns.

### DIFF
--- a/Classes/Domain/Model/Product/Tea.php
+++ b/Classes/Domain/Model/Product/Tea.php
@@ -4,7 +4,7 @@ declare(strict_types=1);
 
 namespace TTN\Tea\Domain\Model\Product;
 
-use TYPO3\CMS\Extbase\Annotation\ORM\Lazy;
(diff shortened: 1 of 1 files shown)


brotkrueml (commented):
<s>Question: The validations are only considered in frontend context, not in backend context with TCA in place. So, what is the need as we don't have a frontend plugin for editing the teas?</s>

Okay, I withdraw the question, just saw #864 . So this is a preparation for editing in FE?

oliverklee:
> So this is a preparation for editing in FE?

It is indeed. 🙂 (I’m trying to split up the FE editing into multiple, self-contained PRs to make the reviews less cumbersome.)

oliverklee:
Background: I’m going to give a workshop soon and am planning to use the Tea extension as an example for CRUD/FE editing.

brotkrueml (approved):

PR #872, opened 2023-06-22
[FEATURE] Add FE editing for existing tea records

oliverklee:
Part of #871

### DIFF
--- a/Classes/Controller/FrontEndEditorController.php
+++ b/Classes/Controller/FrontEndEditorController.php
@@ -5,6 +5,7 @@ declare(strict_types=1);
 namespace TTN\Tea\Controller;
 
 use Psr\Http\Message\ResponseInterface;
+use TTN\Tea\Domain\Model\Product\Tea;
 use TTN\Tea\Domain\Repository\Product\TeaRepository;
 use TYPO3\CMS\Core\Context\Context;
 use TYPO3\CMS\Extbase\Mvc\Controller\ActionController;
@@ -40,5 +41,33 @@ class FrontEndEditorController extends ActionController
     private function getUidOfLoggedInUser(): int
     {
         return $this->context->getPropertyFromAspect('frontend.user', 'id');
+    }
+
+    public function editAction(Tea $tea): ResponseInterface
+    {
+        $this->checkIfUserIsOwner($tea);
+
+        $this->view->assign('tea', $tea);
+
+        return $this->htmlResponse();
+    }
+
+    /**
+     * @throws \RuntimeException
+     */
+    private function checkIfUserIsOwner(Tea $tea): void
+    {
+        if ($tea->getOwnerUid() !== $this->getUidOfLoggedInUser()) {
+            throw new \RuntimeException('You do not have the permissions to edit this tea.', 1687363749);
+        }
+    }
+
+    public function updateAction(Tea $tea): ResponseInterface
+    {
+        $this->checkIfUserIsOwner($tea);
+
+        $this->teaRepository->update($tea);
+
+        return $this->redirect('index');
     }
 }
--- a/Resources/Private/Partials/FrontEndEditor/Form.html
+++ b/Resources/Private/Partials/FrontEndEditor/Form.html
@@ -1,0 +1,36 @@
+<html xmlns:f="http://typo3.org/ns/TYPO3/CMS/Fluid/ViewHelpers" data-namespace-typo3-fluid="true">
+    <f:variable name="propertyLabelPrefix" value="plugin.frontEndEditor.property"/>
+    <f:variable name="idPrefix" value="tea-editor"/>
+
+    <fieldset class="mb-3">
+        <div class="row mb-3">
+            <label for="{idPrefix}-title" class="col-sm-2 col-form-label">
+                <f:translate key="{propertyLabelPrefix}.title"/>
+            </label>
+            <div class="col-sm-10">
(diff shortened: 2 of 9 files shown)


Classes/Controller/FrontEndEditorController.php
```diff
@@ -41,4 +49,33 @@ private function getUidOfLoggedInUser(): int
...
+        return $this->htmlResponse();
+    }
+
+    /**
+     * @throws \RuntimeException
+     */
+    private function checkIfUserIsOwner(Tea $tea): void
+    {
+        if ($tea->getOwnerUid() !== $this->getUidOfLoggedInUser()) {
+            throw new \RuntimeException('You do not have the permissions to edit this tea.', 1687363749);
```
oliverklee, line 68: Note: I'd like to change this to a proper alert in the FE and a "forbidden" HTTP header in a later change (and would like to keep that functionality out of this PR chain here).

oliverklee:
For the redirect, It looks like we'll need to have different code paths depending on the TYPO3 version. On it …

Classes/Controller/FrontEndEditorController.php
```diff
@@ -41,4 +49,33 @@ private function getUidOfLoggedInUser(): int
...
+        }
+    }
+
+    public function updateAction(Tea $tea): ResponseInterface
+    {
+        $this->checkIfUserIsOwner($tea);
+
+        $this->teaRepository->update($tea);
+        $this->persistenceManager->persistAll();
```
DanielSiepmann, line 77: Do we need the `persistAll()`? Shouldn't that be done by Extbase under the hood after the request was handled?
We do a redirect which (at least in earlier versions) finishes the request, in contrast to a forward.

Resources/Private/Templates/FrontEndEditor/Index.html
```diff
@@ -17,6 +17,8 @@ <h2>
                         <th scope="col">
                             <f:translate key="plugin.frontEndEditor.property.title"/>
                         </th>
+                        <th scope="col">
+                        </th>
```
DanielSiepmann, line 21: Should we add a label as well? Something like "Actions"?

Resources/Private/Partials/FrontEndEditor/Form.html
```diff
@@ -0,0 +1,37 @@
+<html xmlns:f="http://typo3.org/ns/TYPO3/CMS/Fluid/ViewHelpers" data-namespace-typo3-fluid="true">
+    <f:variable name="labelPrefix" value="plugin.frontEndEditor"/>
+    <f:variable name="propertyLabelPrefix" value="{labelPrefix}.property"/>
+    <f:variable name="idPrefix" value="tea-editor"/>
```
DanielSiepmann, line 4: I personally am not a fan of the `f:variable` ViewHelper. I don't like to define variables within a View.
I often use the outer render ViewHelper in order to assign the additional context to the inner View. This doesn't work with `_all` but I see `_all` like `$GLOBALS` within PHP and Template, Partials, Sections as methods or functions. That's why I always try to pass the necessary context explicitly.

But that's just a comment, no need to adjust this. But feel free to comment with your own thoughts in order to exchange our different views on that.

Classes/Controller/FrontEndEditorController.php
oliverklee: Indeed (at least in my V11 tests, am I'm not aware of any changes in this area). Done.

Resources/Private/Templates/FrontEndEditor/Index.html
oliverklee: Good idea! Done.

Resources/Private/Partials/FrontEndEditor/Form.html
oliverklee: The rationale behind this variables is to ensure and communicate to the reader that I use the same label prefix, ID prefix, etc. in all places, and to also make those easier to change (as only one place needs to be changed instead of every occurrence).

DanielSiepmann: It sounds like it makes more sense to keep it your way then.

My approach would end in removing the assignments here and altering the render call like this:
```
<f:render partial="FrontEndEditor/Form" arguments="{
    propertyLabelPrefix: 'plugin.frontEndEditor.property',
    idPrefix: 'tea-editor'
}"/>
```
`labelPrefix` was unused, except to define the `propertyLabelPrefix`. That's why it is missing in my example.

But I'll close the comment. Just wanted to share my own thoughts.

oliverklee: Hold on, then I'll inline `labelPrefix`.

(I'd like to keep the variables in the partial instead of in the `f:render` as I'm planning to reuse the partial for the "new tea" form, which would create code duplication with the prefixes again.)

oliverklee: Done.

DanielSiepmann (approved):

DanielSiepmann:
> For the redirect, It looks like we'll need to have different code paths depending on the TYPO3 version. On it …

But I've still approved the current state. Wondering what the difference might be.

oliverklee:
> Wondering what the difference might be.

The difference is that in V11, `redirect()` throws a `StopActionException`, while in V12, it returns a `RedirectResponse`.

PR #876, opened 2023-06-22
[FEATURE] Add a delete functionality for the Tea FE editor

### LINKED ISSUE #871: Add FE editing/CRUD functionality

oliverklee:
The delete action is triggered using a form with a submit button, causing a POST request to be sent instead of a GET request. This is because GET requests should not modify (or delete) data, but only read it and be idempotent. Also, the request then is guaranteed to not get cached.

From a usability perspective, a button instead of a link also is semantically more correct: A link is expected to bring you to some place, whereas a button is expected to trigger some action.

Closes #871

### DIFF
--- a/Classes/Controller/FrontEndEditorController.php
+++ b/Classes/Controller/FrontEndEditorController.php
@@ -9,6 +9,7 @@ use TTN\Tea\Domain\Model\Product\Tea;
 use TTN\Tea\Domain\Repository\Product\TeaRepository;
 use TYPO3\CMS\Core\Context\Context;
 use TYPO3\CMS\Core\Utility\GeneralUtility;
+use TYPO3\CMS\Extbase\Annotation as Extbase;
 use TYPO3\CMS\Extbase\Mvc\Controller\ActionController;
 
 /**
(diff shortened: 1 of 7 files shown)


Classes/Controller/FrontEndEditorController.php
```diff
@@ -89,4 +89,13 @@ public function createAction(Tea $tea): ResponseInterface
 
         return $this->redirect('index');
     }
+
+    public function deleteAction(Tea $tea): ResponseInterface
+    {
+        $this->checkIfUserIsOwner($tea);
+
+        $this->teaRepository->remove($tea);
+
```
DanielSiepmann, line 98: I guess you'll add a FlashMessage with another PR?

Resources/Private/Templates/FrontEndEditor/Index.html
```diff
@@ -40,6 +40,12 @@ <h2>
                                     <f:translate key="plugin.frontEndEditor.action.edit"/>
                                 </f:link.action>
                             </td>
+                            <td>
+                                <f:form action="delete" name="tea" object="{tea}">
```
DanielSiepmann, line 44: I can't remember when I did this the last time. So to be sure: Extbase/Fluid will add a hidden input with the uid of the record only by adding the object to the form? Or do we need to add that input field ourself?

DanielSiepmann (approved):

Classes/Controller/FrontEndEditorController.php
oliverklee: Yes, exactly. I added #875 to track this.

Resources/Private/Templates/FrontEndEditor/Index.html
oliverklee: Exactly. In 11LTS, there 7 hidden fields get auto-added. One of them is for the identity:

```php
<input type="hidden" name="tx_tea_teafrontendeditor[tea][__identity]" value="2">
```

Issue #830, opened 2023-05-03
Implement type coverage calculation


lukaszuznanski:
We can ensure that our code is using types and calculate percentage of type coverage, we could use this tool
https://github.com/TomasVotruba/type-coverage

PR #889, opened 2023-07-10
Add type coverage

lukaszuznanski:
Resolves: #830

### DIFF
--- a/phpstan.neon
+++ b/phpstan.neon
@@ -22,6 +22,6 @@ parameters:
     - Tests
 
   type_coverage:
-      return_type: 50
-      param_type: 30
-      property_type: 70
+      return_type: 100
+      param_type: 100
+      property_type: 100
--- a/phpstan.neon
+++ b/phpstan.neon
@@ -24,4 +24,4 @@ parameters:
   type_coverage:
       return_type: 100
       param_type: 100
-      property_type: 100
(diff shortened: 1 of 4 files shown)


phpstan.neon
```diff
@@ -20,3 +20,8 @@ parameters:
     - Classes
     - Configuration
     - Tests
+
+  type_coverage:
+      return_type: 50
+      param_type: 30
+      property_type: 70
```
oliverklee, line 27: Let's go a lot higher!

lukaszuznanski:
Damn, I clicked on update branch and merged main into branch... sorry, let's check if there is way to disable this button, or change it's behavior. I will fix this error in functional tests.

oliverklee:
I've fixed this now:

```bash
git switch Feature/Add-type-coverage
git reset --hard a14f676
git rebase
git push -f
```

lukaszuznanski:
I changed it to 100, 100 and 95 for properties. We can update it to 100, but first we need to add property type. This should work for now. Let's do it in another PR.

oliverklee (approved):

oliverklee (changes requested):
Looks good. It should get an entry in the changelog, though.

oliverklee (approved):

PR #904, opened 2023-08-01
[REGRESSION] Use ghcr docs render image

### LINKED ISSUE #903: [REGRESSION] Use ghcr docs render image
The new runTests.sh relies on the ghcr.io image. However the image itself relies on a non-prefixed version of the image. Thus, the image can't be pulled automatically, because it needs to be tagged afterwards.

Additionally the composer.json currently refers to the non-ghcr.io prefixed image, which isn't existing anymore.

This solution is based on QuickStart Step 1 of https://typo3.slack.com/archives/C028JEPJL/p1686763606638419

I don't know if calling it before every `composer run docs:generate` is necessary, but as

1. EXT:tea currently doesn't offer an "init" script to init the dev environment (e.g. `composer i && yarn`) and
2. The error message without the tagging: `Unable to find image 't3docs/render-documentation:latest' locally
docker: Error response from daemon: pull access denied for t3docs/render-documentation, repository does not exist or may require 'docker login': denied: requested access to the resource is denied.` is not self explanatory

Thus, I propose this rather rigid approach to run the pull every time.

EDIT: to ease discussion here is the excerpt from the commit c1723dd with the two commands necessary to run to init the local docker environment, pulling the image and creating an alias for the non-prefixed version via `docker tag`:
```	
docker pull ghcr.io/t3docs/render-documentation
docker tag ghcr.io/t3docs/render-documentation t3docs/render-documentation
```

jpmschuler:
The new runTests.sh relies on the ghcr.io image. However the image itself relies on a non-prefixed version of the image. Thus, the image can't be pulled automatically, because it needs to be tagged afterwards.

fixes #903

### DIFF
--- a/composer.json
+++ b/composer.json
@@ -154,7 +154,12 @@
 		"ci:yaml:lint": "find . ! -path '*.Build/*' ! -path '*node_modules/*' -regextype egrep -regex '.*.ya?ml$' | xargs -r php ./.Build/bin/yaml-lint",
 		"coverage:create-directories": "mkdir -p .Build/logs .Build/coverage",
 		"docs:generate": [
-			"docker run --rm t3docs/render-documentation show-shell-commands > tempfile.sh; echo 'dockrun_t3rd makehtml' >> tempfile.sh; bash tempfile.sh; rm tempfile.sh"
+			"@docs:generate:pullimage",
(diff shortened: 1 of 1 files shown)


oliverklee (approved):
Looks good. Thanks for fixing this!

Issue #902, opened 2023-07-31
Use our own copies of the Testing Framework configuration files


oliverklee:
This is what the testing framework (strongly) recommends.

PR #952, opened 2023-10-17
[TASK] Use own configuration files for tests

lukaszuznanski:
Resolves: #902

### DIFF
--- a/Tests/Functional/FunctionalTests.xml
+++ b/Tests/Functional/FunctionalTests.xml
@@ -1,0 +1,42 @@
+<phpunit
+    backupGlobals="true"
+    bootstrap="../../.Build/vendor/typo3/testing-framework/Resources/Core/Build/FunctionalTestsBootstrap.php"
+    cacheResult="false"
+    colors="true"
+    convertErrorsToExceptions="true"
+    convertWarningsToExceptions="true"
+    convertDeprecationsToExceptions="true"
+    convertNoticesToExceptions="true"
+    forceCoversAnnotation="false"
+    stopOnError="false"
+    stopOnFailure="false"
+    stopOnIncomplete="false"
+    stopOnSkipped="false"
+    verbose="false"
+    beStrictAboutTestsThatDoNotTestAnything="false"
+    failOnWarning="true"
+    failOnRisky="true"
+>
+    <testsuites>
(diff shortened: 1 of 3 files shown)


Tests/Functional/FunctionalTests.xml
```diff
@@ -0,0 +1,42 @@
+<phpunit
```
oliverklee, line 1: AFAIK, the PHPUnit configuration (at least as written by PHPUnit) contains the XML prologue and the schema information:

```xml
<?xml version="1.0" encoding="UTF-8"?>
<phpunit xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance"
         xsi:noNamespaceSchemaLocation="https://schema.phpunit.de/9.6/phpunit.xsd"
```

oliverklee (changes requested):

Tests/Unit/UnitTests.xml
```diff
@@ -0,0 +1,49 @@
...
+            <!--
+                This path either needs an adaption in extensions, or an extension's
+                test location path needs to be given to phpunit.
+            -->
+            <directory suffix="Test.php">./</directory>
+        </testsuite>
+    </testsuites>
+    <!-- @todo: change tag to 'coverage' when TF requires phpunit > 9 -->
+    <filter>
+        <!-- @todo: change tag to 'include' when TF requires phpunit > 9 -->
```
oliverklee, line 35: We're already using PHPUnit 9. So we should witch to `coverage` and `include` here.

oliverklee (changes requested):

oliverklee (approved):

sbuerk:
@oliverklee @lukaszuznanski 

Little bit late. TBH the bootstraps should also not be used from the the `typo3/testing-framework` vendor folder and copied to the project.

The background is, that there are scenarious where it's needed to make some changes based on the project/extension.

Beside this, the chosen location is kind of flattered. Configuration for all tools should be collected in one place and not distributed throughout a lot of places (and not the root folder)
to have a clean and "one place" to look it up. But that is maybe more a personally flavour - where I highly disagree to take this places as `best practice`.

oliverklee:
Thanks, I've created #1082 for moving these files.

PR #962, opened 2023-11-01
[TASK] Use more stubs in the unit tests

### LINKED ISSUE #870: Use `createStub` in the unit tests more often
This is for mocks that should not get their constructor called, and that should not retain any of their original functionality.

oliverklee:
This way, the purpose of the stub/mock is more explicit.

Fixes #870

### DIFF
--- a/Tests/Unit/Controller/TeaControllerTest.php
+++ b/Tests/Unit/Controller/TeaControllerTest.php
@@ -51,8 +51,8 @@ final class TeaControllerTest extends UnitTestCase
         $this->viewMock = $this->createMock(TemplateView::class);
         $this->subject->_set('view', $this->viewMock);
 
-        $responseMock = $this->createMock(HtmlResponse::class);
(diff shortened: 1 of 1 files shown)


Tests/Unit/Controller/TeaControllerTest.php
```diff
@@ -51,7 +51,7 @@ protected function setUp(): void
         $this->viewMock = $this->createMock(TemplateView::class);
         $this->subject->_set('view', $this->viewMock);
 
-        $responseMock = $this->createMock(HtmlResponse::class);
+        $responseMock = $this->createStub(HtmlResponse::class);
```
DanielSiepmann, line 54: Shouldn't we also rename the variable from Mock to Stub?

oliverklee: Good catch! Done.

DanielSiepmann (approved):

PR #989, opened 2023-11-16
[FEATURE] Make records timeable and hidable

lukaszuznanski:
Resolves: #553 

Tested on v12 and v11

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -6,11 +6,11 @@ This project adheres to [Semantic Versioning](https://semver.org/).
 ## x.y.z
 
 ### Added
+- Make records timeable and hideable (#989)
 - Add support for PHP 8.3 (#965)
 - Add an FE editor (#864, #872, #874, #876)
 - Add automerging of green Dependabot PRs (#756)
 - Add type coverage calculation (#830)
-- Make records timeable and hideable (#553)
 
 ### Changed
 - Rename TsConfig directory to `TSconfig` (#923)
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -10,6 +10,7 @@ This project adheres to [Semantic Versioning](https://semver.org/).
 - Add an FE editor (#864, #872, #874, #876)
 - Add automerging of green Dependabot PRs (#756)
 - Add type coverage calculation (#830)
+- Make records timeable and hideable (#553)
 
 ### Changed
 - Rename TsConfig directory to `TSconfig` (#923)
--- a/Configuration/TCA/tx_tea_domain_model_product_tea.php
+++ b/Configuration/TCA/tx_tea_domain_model_product_tea.php
@@ -245,11 +245,13 @@ if ($typo3Version->getMajorVersion() < 12) {
     ];
     $tca['columns']['hidden']['config'] = [
         'type' => 'check',
+        'label' => 'LLL:EXT:core/Resources/Private/Language/locallang_general.xlf:LGL.enabled',
         'items' => [
-            0 => '',
-            1 => '',
+            [
+                0 => '',
+                'invertStateDisplay' => true,
+            ],
         ],
-        'invertStateDisplay' => true,
     ];
     $tca['columns']['starttime']['config'] = [
         'type' => 'input',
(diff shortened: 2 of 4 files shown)


CHANGELOG.md
```diff
@@ -10,6 +10,7 @@ This project adheres to [Semantic Versioning](https://semver.org/).
 - Add an FE editor (#864, #872, #874, #876)
 - Add automerging of green Dependabot PRs (#756)
 - Add type coverage calculation (#830)
+- Make records timeable and hideable (#553)
```
oliverklee, line 13: This line should go on top so the changelog is in reverse chronological order.

Configuration/TCA/tx_tea_domain_model_product_tea.php
```diff
@@ -24,16 +27,64 @@
...
+                endtime;LLL:EXT:frontend/Resources/Private/Language/locallang_ttc.xlf:endtime_formlabel,
+                --linebreak--,
+                fe_group;LLL:EXT:frontend/Resources/Private/Language/locallang_ttc.xlf:fe_group_formlabel,
+            ',
         ],
     ],
     'columns' => [
+        'hidden' => [
+            'exclude' => true,
+            'label' => 'LLL:EXT:core/Resources/Private/Language/locallang_general.xlf:LGL.enabled',
```
oliverklee, line 53: This label doesn't quit match with teas: In German, the labels says "Sichtbarkeit des Inhaltelemens" (translated: "visibility of the content element"). As tea records are no content elements, we probably will need to have our own label for this.

oliverklee (changes requested):
I've manually tested this in a 11LTS and 12LTS backend: It works fine for me in both version. Thanks!

We should change a label, though.

CHANGELOG.md
oliverklee: Oh, and also we should use the new PR number #989.

Configuration/TCA/tx_tea_domain_model_product_tea.php
oliverklee: My proposal for the label:
- "Visible" (EN)
- "Sichtbar" (DE)

Resources/Private/Language/de.locallang_db.xlf
```diff
@@ -31,6 +31,9 @@
 				<source>Usergroup Access Rights</source>
 				<target>Zugriffsrechte für Benutzergruppen</target>
 			</trans-unit>
+			<trans-unit id="tx_tea_domain_model_product_tea.hidden">
+				<source>Sichtbar</source>
```
oliverklee, line 35: This needs to be the `target` (with the `source` copied from the `locallang_db.xml`).

oliverklee (changes requested):

Resources/Private/Language/de.locallang_db.xlf
lukaszuznanski: Yea, my fault, fixing right now.

oliverklee:
Hm, this is strange. With the latest version (on another machine, though), I still see the "visibility of content element" label in V11, and don't see the new fields in the BE in V12 at all.

Configuration/TCA/tx_tea_domain_model_product_tea.php
```diff
@@ -24,16 +27,64 @@
...
+            'l10n_mode' => 'exclude',
+            'l10n_display' => 'defaultAsReadonly',
+        ],
+        'endtime' => [
+            'exclude' => true,
+            'label' => 'LLL:EXT:core/Resources/Private/Language/locallang_general.xlf:LGL.endtime',
+            'config' => [
+                'type' => 'datetime',
+                'default' => 0,
+                'range' => [
```
oliverklee, line 81: Why is there an upper limit for the endtime, but not for the starttime?

lukaszuznanski: I've taken it from tt_content from core, should we adjust it and also fix it for core?

oliverklee (approved):
Works fine on my V11 and V12 installations. Thanks!

PR #991, opened 2023-11-16
[TASK] Upgrade DDEV

lukaszuznanski:


.ddev/config.yaml
```diff
@@ -1,33 +1,31 @@
 name: tea
 type: php
 docroot: ""
-php_version: "8.1"
+php_version: "8.3"
```
oliverklee, line 4: Please let's move the PHP upgrade to a separate PR. (Also, we need to stay below PHP 8.3 in order to not break PHP-CS-Fixer at this moment.)

.ddev/config.yaml
```diff
@@ -1,33 +1,31 @@
...
-mutagen_enabled: false
+    type: mariadb
+    version: "10.5"
 hooks:
-  post-start:
-  - exec: composer install --no-progress
-omit_containers: [dba, ddev-ssh-agent]
+    post-start:
+        - exec: composer install --no-progress
+omit_containers: [ddev-ssh-agent]
```
oliverklee, line 17: Shouldn't we still omit the DBA container? Or is this related to the DDEV upgrade?

.ddev/config.yaml
```diff
@@ -1,33 +1,31 @@
...
 xdebug_enabled: false
 additional_hostnames: []
 additional_fqdns: []
 database:
-  type: mariadb
-  version: "10.5"
-nfs_mount_enabled: false
-mutagen_enabled: false
+    type: mariadb
+    version: "10.5"
```
oliverklee, line 13: I'd prefer to have the formatting cleanup in a separate PR as a pre-change in order to separate the formatting changes from the real configuration changes.

oliverklee (changes requested):

.ddev/config.yaml
lukaszuznanski: Those changes comes from ddev upgrade, it automatically reformat this file

.ddev/config.yaml
lukaszuznanski: Will fix, related to ddev upgrade, but it should be migrated also.

.ddev/config.yaml
lukaszuznanski: Ok

oliverklee (approved):

PR #1020, opened 2023-11-27
[FEATURE] Provide Polish translation for tea

lukaszuznanski:
Resolves: #1014


oliverklee (changes requested):
Looks good (from a technical standpoint; I don't speak Polish).

Please let's also have a changelog entry for this.

Issue #859, opened 2023-06-20
Add first functional test using request as showcase


DanielSiepmann:
The goal is to demonstrate how requests can be made within functional tests in order to test the rendered result.

lukaszuznanski:
I have example setup, maybe this will be useful in this issue
https://github.com/TYPO3-Headless/headless/blob/master/Tests/Functional/BaseTest.php
https://github.com/TYPO3-Headless/headless/blob/master/Tests/Functional/ContentTypes/ImageElementTest.php

oliverklee:
More examples:
https://github.com/werkraum-media/calendar/blob/main/Tests/Functional/CalendarControllerTest.php
https://github.com/werkraum-media/events/blob/main/Tests/Functional/Frontend/DatesTest.php
https://github.com/werkraum-media/watchlist/blob/main/Tests/Functional/BasicsTest.php
https://github.com/werkraum-media/thuecat/blob/main/Tests/Functional/FrontendTest.php

PR #1021, opened 2023-11-27
[FEATURE] Add first functional test using a request

DanielSiepmann:
That way the extension serves as an example on how to use the TYPO3 internal requests provided by the TYPO3 testing framework. Those can be used as integration tests.
They are also often easier to set up on existing projects and allow to refactor the code base, compared to functional and unit tests.

Resolves: #859


Tests/Functional/Frontend/AbstractTest.php
```diff
@@ -0,0 +1,34 @@
...
+declare(strict_types=1);
+
+namespace TTN\Tea\Tests\Functional\Frontend;
+
+use TYPO3\TestingFramework\Core\Functional\FunctionalTestCase;
+
+/**
+ * Provides general set up for frontend related functional tests.
+ */
+abstract class AbstractTest extends FunctionalTestCase
```
oliverklee, line 12: As far as I understand it, this is the base class for the controller tests. So I suggest the name `AbstractControllerTest`, and move it to `Tests/Functional/Controller`.

Tests/Functional/Frontend/TeaIndexTest.php
```diff
@@ -0,0 +1,32 @@
...
+
+namespace TTN\Tea\Tests\Functional\Frontend;
+
+use TTN\Tea\Frontend\TeaIndex;
+use TYPO3\TestingFramework\Core\Functional\Framework\Frontend\InternalRequest;
+
+/**
+ * @coversNothing
+ */
+final class TeaIndexTest extends AbstractTest
```
oliverklee, line 13: I'd suggest making this test about the whole controller, i.e., rename it to `TeaControllerTest`, move it to `Tests/Functional/Controller` and have it `@covers` the controller.

Tests/Functional/Frontend/TeaIndexTest.php
```diff
@@ -0,0 +1,32 @@
...
+{
+    /**
+     * @test
+     */
+    public function rendersAllAvailableTeas(): void
```
oliverklee, line 18:
```suggestion
    public function indexActionRendersAllAvailableTeas(): void
```

Tests/Functional/Frontend/TeaIndexTest.php
```diff
@@ -0,0 +1,32 @@
...
+        $this->importCSVDataSet(__DIR__ . '/Fixtures/Database/Teas.csv');
+
+        $request = new InternalRequest();
+        $request = $request->withPageId(1);
+
+        $html = $this->executeFrontendRequest($request)->getBody()->__toString();
+
+        self::assertStringContainsString('Godesberger Burgtee', $html);
+        self::assertStringContainsString('Earl Grey', $html);
+        self::assertStringContainsString('Oolong', $html);
```
oliverklee, line 30: For this testcase to be minimal, two teas should suffice (in the test data and the assertions).

oliverklee (changes requested):
PHP-CS-Fixer also is unhappy. ;-)

Tests/Functional/Frontend/AbstractTest.php
DanielSiepmann: Well it is related for all frontend related stuff. We have some extensions where we provide ViewHelper, UserFunc, Middleware or other stuff not necessarily related to controller. Controller are a bit to "deep knowledge" regarding testing a request for my taste.
That's why I used 'Frontend' as folder and namespace.

But I'm fine to adjust this, especially as I don't see that EXT:tea will get anything like that soon.

oliverklee (approved):

PR #1023, opened 2023-11-27
Added Dutch translation

frschep:


lukaszuznanski (commented):
I think that you should leave <target> in every file in order to make this work correctly.

Resources/Private/Language/nl.locallang.xlf
```diff
@@ -0,0 +1,65 @@
+<?xml version="1.0" encoding="UTF-8" standalone="yes" ?>
+<xliff xmlns="urn:oasis:names:tc:xliff:document:1.2" version="1.2">
+	<file source-language="en" datatype="plaintext" original="messages">
+		<header/>
+		<body>
+			<trans-unit id="plugin.tea_index">
+				<source>Theelijst</source>
```
oliverklee, line 7: The NL localization needs to be in the  `<target>` nodes, while the original EN text needs to be in the `<source>` nodes.

Also, the target language needs to be referenced in the file header. Please have a look at the DE and PL localization files for this.

oliverklee (changes requested):
Congratulations on your first PR!

Could you also please add a changelog entry for this (newest on top, using the PR ID)? Thanks!

PR #1025, opened 2023-11-27
[!!!][TASK] Drop additional namespace segment for the Tea model

### LINKED ISSUE #1008: Drop the additional "Product" namespace from the models
This currently makes understanding the project structure harder.

oliverklee:
The `Product` namespace segment in the domain model namespace `TTN\Tea\Domain\Model` currently serves no purpose and only adds confusion. So let's simplify the extension structure accordingly.

(I intended to use this to demonstrate DDD contexts, but never built enough models in the Tea extension for this to actually make sense.)

Fixes #1008

### DIFF
--- a/Tests/Unit/Controller/FrontEndEditorControllerTest.php
+++ b/Tests/Unit/Controller/FrontEndEditorControllerTest.php
@@ -6,8 +6,8 @@ namespace TTN\Tea\Tests\Unit\Controller;
 
 use PHPUnit\Framework\MockObject\MockObject;
 use TTN\Tea\Controller\FrontEndEditorController;
-use TTN\Tea\Domain\Model\Product\Tea;
-use TTN\Tea\Domain\Repository\Product\TeaRepository;
+use TTN\Tea\Domain\Model\Tea;
+use TTN\Tea\Domain\Repository\TeaRepository;
 use TYPO3\CMS\Core\Context\Context;
 use TYPO3\CMS\Core\Context\UserAspect;
 use TYPO3\CMS\Core\Http\HtmlResponse;
--- a/Tests/Unit/Domain/Repository/TeaRepositoryTest.php
+++ b/Tests/Unit/Domain/Repository/TeaRepositoryTest.php
@@ -2,15 +2,15 @@
 
 declare(strict_types=1);
 
-namespace TTN\Tea\Tests\Unit\Domain\Repository\Product;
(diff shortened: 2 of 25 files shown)


Tests/Unit/Domain/Repository/TeaRepositoryTest.php
```diff
@@ -25,7 +25,7 @@ protected function setUp(): void
             // @phpstan-ignore-next-line This line is 11LTS-specific, but we're running PHPStan on TYPO3 12.
             $this->subject = new TeaRepository($objectManagerStub);
         } else {
-            $this->subject = new TeaRepository();
+            $this->subject = new \TTN\Tea\Domain\Repository\TeaRepository();
```
DanielSiepmann, line 28: Why do we introduce the FCQN for this line while using the imported namespace on line 26?

Tests/Unit/Controller/FrontEndEditorControllerTest.php
```diff
@@ -39,7 +39,7 @@ final class FrontEndEditorControllerTest extends UnitTestCase
     private Context $context;
 
     /**
-     * @var TeaRepository&MockObject
+     * @var \TTN\Tea\Domain\Repository\TeaRepository&MockObject
```
DanielSiepmann, line 42: Why do we introduce the FQCN here?

DanielSiepmann (commented):
I'd only question whether we want a breaking change right now.
Should we plan breaking changes together with major version releases?

Tests/Unit/Domain/Repository/TeaRepositoryTest.php
oliverklee: Done.

Tests/Unit/Controller/FrontEndEditorControllerTest.php
oliverklee: Done.

oliverklee:
@DanielSiepmann @lukaszuznanski ping for a review - thanks!

lukaszuznanski (approved):
Should we add any migration guide?

DanielSiepmann:
Data is not affected anyway, so not a showcase for how to add UpgradeWizards anyway. This would only be an opportunity to provide a rector rule. But I also won't think we should do that right now.

DanielSiepmann (approved):

Issue #1024, opened 2023-11-27
Make the repository use the storage PID


oliverklee:
Basically, stop using the `StoragePageAgnosticTrait` and drop it.

Also document that the storage PID needs to be set for the list view to work.

`TeaRepository::findByOwnerUid` needs to be adapted to ignore the storage PID (as it should find records from all pages).

Starkmann:
Storage pid is set via TS and already part of the existing fixutre

Starkmann:
findAllWithRecordsFindsRecordsFromAllPages this Test does not make scene any more, but  im not sure how/if we want to test if storage pid is respected if set.

PR #1028, opened 2023-11-27
[TASK] remove trait and respect storage pid

Starkmann:
Ref:#1024

Storage pid is set via TS and already part of the existing fixture


oliverklee:
I'm adopting this change in #1223 (if that's okay with you, @Starkmann).

oliverklee:
Done in #1223 now. Closing.

Issue #976, opened 2023-11-02
Fix functional tests on gitlab


lukaszuznanski:
functional tests are failing https://git.typo3.org/qa/example-extension/-/jobs/2694974

oliverklee:
We might want to bisect the minimum TYPO3 12 version we need to get this running (also to find out if this is related to a fixed bug in the Core).

oliverklee:
Some V11 GitLab jobs are installing V12 packages:
- func-php8.1-v11-highest 
- func-php8.1-v11-lowest 
- func-php8.2-v11-highest 
- func-php8.2-v11-lowest 

I'm looking into it.

oliverklee:
Latest GitLab job for `main`: https://git.typo3.org/qa/example-extension/-/pipelines/49990

lukaszuznanski:
Can confirm on local environment by using those steps from ```yml``` configuration file.

PR #1043, opened 2023-12-04
Fix for functional tests on gitlab CI

lukaszuznanski:
Resolves: #976 

This is related to different settings in error reporting.
https://git.typo3.org/qa/example-extension/-/commit/ccc186123268e8fc0af2cb3d6447d9eed8bc3394


oliverklee:
Can't we set the INI values via `Tests/Functional/FunctionalTests.xml` instead?

lukaszuznanski:
> Can't we set the INI values via `Tests/Functional/FunctionalTests.xml` instead?

This sounds like much better solution. I will.

lukaszuznanski:
It looks like it doesn't work properly when it's set in phpunit.xml, maybe it's too late to set this?

oliverklee:
Okay, then let's keep the change as is.

oliverklee (approved):

PR #1052, opened 2023-12-09
Remove leftover select_key

nhovratov:
select_key does not exist anymore in the Core and can be safely removed,


oliverklee (approved):
Thanks! :heart:

PR #1058, opened 2023-12-12
[BUGFIX] Create required directories before running the functional tests

### LINKED ISSUE #1053: Create directory before running the functional tests on CI
https://github.com/FriendsOfTYPO3/tea/actions/runs/7153662905/job/19481602763?pr=1052

```

Run export typo3DatabaseName="$DB_DATABASE";
> find 'Tests/Functional' -wholename '*Test.php' | parallel --gnu 'echo; echo "Running functional test suite {}"; .Build/bin/phpunit -c ./Tests/Functional/FunctionalTests.xml {}';

Running functional test suite Tests/Functional/Domain/Repository/Product/TeaRepositoryTest.php
PHPUnit 9.6.15 by Sebastian Bergmann and contributors.

Error in bootstrap script: TYPO3\TestingFramework\Core\Exception:
Directory "/home/runner/work/tea/tea/.Build/Web/typo3temp/var/tests" could not be created
```

This seems to be some timing-related issue or race condition that makes the functional tests flaky on GitHub Actions.

We should create the directory before running the tests.

oliverklee:
This prevents the occasional "directory could not be created" failure when running the functional tests on CI.

Fixes #1053

### DIFF
--- a/composer.json
+++ b/composer.json
@@ -119,6 +119,7 @@
 			"@ci:coverage:functional"
 		],
 		"ci:coverage:functional": [
+			"@ci:tests:create-directories",
 			"@coverage:create-directories",
 			".Build/bin/phpunit -c ./Tests/Functional/FunctionalTests.xml --whitelist Classes --coverage-php=.Build/coverage/functional.cov Tests/Functional"
 		],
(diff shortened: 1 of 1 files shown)


DanielSiepmann (approved):
Fine for me. Yet I never had that issue.

oliverklee:
As far as I can tell, this is a timing-related issue when the tests are run in parallel (which usually takes place in CI). So the functional tests actually *do* try to create the directory (somewhere in their bootstrap), but end up blocking each other. This happens on CI quite often. I had to rerun the CI pipeline a couple of days during the last few days because of this.

I'd prefer to not have to set up our own tests bootstrapping for this (that then would need to copy or run the bootstrap from the testing framework).

DanielSiepmann:
Thanks for the explanation :) We use https://packagist.org/packages/brianium/paratest in the single project where we parallise the execution of functional tests. Maybe it works a bit different and prevents the issue. Or maybe we didn't encounter them as there is a difference with GitLab.

Issue #1087, opened 2023-12-19
[BUG] `podman` does not support the `uid` argument for a tmpfs volume mount


sbuerk:
The TYPO3 core `runTests.sh` has been using the
`uid` option for the `tmpfs` mount in the past
and removed it due to the fact it's not supported
by the `podman` container binary.

This has been missed to change along with the
introduction of the `-b` container binary option
and therefore the podmand and docker dual support.

For example, following

```shell
--tmpfs ${ROOT_DIR}/typo3temp/var/tests/functional-sqlite-dbs/:rw,noexec,nosuid,uid=${HOST_UID}
```

should be changed to

```shell
--tmpfs ${ROOT_DIR}/typo3temp/var/tests/functional-sqlite-dbs/:rw,noexec,nosuid
```

which works for both binaries in case of tmpfs mounts

oliverklee:
Possibly a duplicate of or related to #1083?

sbuerk:
yes - not saw that issue, sorry for the double issue.

Issue #1083, opened 2023-12-19
Invalid mount option when running the functional tests with Podman


oliverklee:
This is with Podman 4.3.1.

```
klee@gonzales ~/src/typo3/ext/tea (main)$ ./Build/Scripts/runTests.sh -s functional -b podman
Error: unknown mount option "uid=1000": invalid mount option

###########################################################################
Result of functional
Environment: local
PHP: 8.1
TYPO3: 11.5
CONTAINER_BIN: podman
DBMS: sqlite
FAILURE
###########################################################################
```

@sbuerk

Issue #1075, opened 2023-12-17
[TASK] Adopt `-b` option to select the container binard for `runTests.sh`


sbuerk:
TYPO3 core introduced the `-b` option for the
`Build/Scripts/runTests.sh` script to provide
the ability to select the container binary,
`docker` or `podman` and enforces podman for
CI (GitLab runner) runs after switching the
Core CI infrastructure from Docker-In-Docker
to Podman-In-Podman.

The `ext:tea` adoption of the `runTests.sh`
has been already adopted on a intermediate
version and dropped the `docker-compose`
based container definitions. Along with
other minor prepration patches this allows
now to add the container binary selection
also.

Subtask of #969

PR #1089, opened 2023-12-19
[BUGFIX] Avoid unsupported option for `podman`

sbuerk:
The TYPO3 core runTests.sh has been using the
uid option for the tmpfs mount in the past
and removed it due to the fact it's not supported
by the podman container binary.

This has been missed to change along with the
introduction of the `-b` container binary option
in #1075 and therefore the podmand and docker dual
support.

For example, following

```shell
--tmpfs ${ROOT_DIR}/typo3temp/var/tests/functional-sqlite-dbs/:rw,noexec,nosuid,uid=${HOST_UID}
```

is changed to

```shell
--tmpfs ${ROOT_DIR}/typo3temp/var/tests/functional-sqlite-dbs/:rw,noexec,nosuid
```

which works for both binaries in case of tmpfs mounts.

Resolves: #1087
Resolves: #1083
Related: #1075


oliverklee (approved):
Thanks for the quick fix!

PR #1092, opened 2023-12-19
[TASK] Add coverage configuration for functional tests

oliverklee:
We already had this for unit tests, but it was missing for functional tests.

### DIFF
--- a/Tests/Functional/FunctionalTests.xml
+++ b/Tests/Functional/FunctionalTests.xml
@@ -20,6 +20,7 @@
     stopOnSkipped="false"
     verbose="false"
 >
+    <coverage/>
     <testsuites>
         <testsuite name="Functional tests">
             <!--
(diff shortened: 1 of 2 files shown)


Tests/Functional/FunctionalTests.xml
```diff
@@ -29,6 +29,14 @@
             <directory suffix="Test.php">./</directory>
         </testsuite>
     </testsuites>
+    <coverage>
+        <include>
+            <!--
+                This path needs an adaption in extensions, when coverage statistics are wanted.
+            -->
+            <directory>../../../../../../typo3/sysext/*/Classes/</directory>
```
DanielSiepmann, line 37: I see wee have the same path for unit tests.
Maybe I don't understand how this works, but I would expect we need to define the Classes folder of EXT:tea instead of foreign code.
Can you explain why we include the system extensions instead of our own Classes folder?

oliverklee: I've done some more research: It turns out we don't need to provide any paths here - having an empty `coverage` node is enough to make PHPUnit happy with the configuration file.

lukaszuznanski (approved):

PR #1108, opened 2024-01-05
[TASK] Move the PHPUnit configuration files to `Configuration/`

### LINKED ISSUE #1082: Move the PHPUnit configuration files to `Configuration/`
We also need to update the `.gitattributes` file and the `prepare-release` Composer Scripts to keep the files out of the releases.

oliverklee:
The `Tests/` directory should only include test code, but not the configuration files.

Fixes #1082

### DIFF
--- a/.gitattributes
+++ b/.gitattributes
@@ -11,6 +11,8 @@
 /.phpstorm.meta.php export-ignore
 /.prettierrc.js export-ignore
 /Build/ export-ignore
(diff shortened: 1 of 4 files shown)


oliverklee:
Note: The build failure is unrelated to my changes. It's caused by a [problem in PHPStan 1.10.52](https://github.com/phpstan/phpstan/issues/10387).

lukaszuznanski (approved):

PR #1116, opened 2024-01-13
[TASK] Migrate the deprecated `executeFrontendRequest` in tests

oliverklee:
This method is deprecated. Starting with TYPO3 11LTS, `executeFrontendSubRequest` should be used.


DanielSiepmann (approved):

DanielSiepmann:
I worked way too much with older versions where this was not usable yet. Sorry for introducing the deprecation method call.

oliverklee:
> Sorry for introducing the deprecation method call.

No worries - this happens to me all the time as well. And learning about the changes to the best practices is one of the reasons why we do this, right? :slightly_smiling_face:

PR #1136, opened 2024-01-24
[TASK] Prefer string casts over explictly calling `__toString`

oliverklee:


DanielSiepmann:
@oliverklee can you explain the reason?
I actually do it the other way around, as this explicitly communicates this is an object which can be casted.

oliverklee:
> can you explain the reason?

Some static analysis tool with the default configuration complained about this in one of my projects. (Unfortunately, I don't remember which one it was.)

DanielSiepmann (approved):

DanielSiepmann:
Okay, thanks :) Was curious. But it is not too important for myself.

Issue #1140, opened 2024-01-26
Use `public` folder instead of `Web`


simonschaufi:
`Web` comes from the Neos project where it is still used until today but in TYPO3 the `Web` folder for public files is not common. Therefore it should be changed to `public`.

oliverklee:
Thanks for bringing this up, @simonschaufi!

PR #1144, opened 2024-01-30
[TASK] Change webdir of TYPO3 to public

lukaszuznanski:
Resolves: #1140

It's blocked right now as it's breaking change.


oliverklee:
> It's blocked right now as it's breaking change.

Is is breaking anything else or blocked by anything else?

Apart from that, we're currently on our way to version 4.0 and already merged the first breaking change in #1025. So let's have these breaking changes! ;-)

oliverklee (approved):

PR #1151, opened 2024-02-03
[TASK] Add PHP CS Fixer to GitLab

simonschaufi:
I've just noticed that PHP CS Fixer is only executed on GitHub but not on GitLab


simonschaufi:
I've just noticed that PHP CS Fixer is only executed on GitHub but not on GitLab

oliverklee (approved):
Thanks, looks good!

I'd like to wait for the build to appear on https://git.typo3.org/qa/example-extension/-/pipelines so we can see this in action before merging.

oliverklee:
Ah, we won't see this on GitLab for fork. Then I'll merge right now.

Issue #1158, opened 2024-02-06
 Add PHPStan extension for disallowed calls


DanielSiepmann:
We decided to add this PHPStan extension: https://github.com/spaze/phpstan-disallowed-calls

PR #1159, opened 2024-02-06
[TASK] Add PHPStan extension for disallowed calls

DanielSiepmann:
A new composer package "spaze/phpstan-disallowed-calls" is added. This one is a PHPStan extension which will check for forbidden calls. Those calls need to be defined per project.
We use this to prevent debugging and other none good practices from showing up in our code base.

We include the pre defined rules from the package. Those can be argued and adjusted once we hit them.

Resolves: #1158

### DIFF
--- a/composer.json
+++ b/composer.json
@@ -58,6 +58,7 @@
 		"phpunit/phpunit": "^9.6.16",
 		"saschaegerer/phpstan-typo3": "^1.10.0",
 		"seld/jsonlint": "^1.10.1",
+		"spaze/phpstan-disallowed-calls": "^3.1",
 		"squizlabs/php_codesniffer": "^3.8.1",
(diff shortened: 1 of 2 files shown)


DanielSiepmann:
Feel free to add your own ideas.

DanielSiepmann:
The result currently would look like this:
```
 ------ ---------------------------------------------------------------------------------------------------------------------------------------------
  Line   Classes/Controller/TeaController.php
 ------ ---------------------------------------------------------------------------------------------------------------------------------------------
  26     Using $_GET is forbidden, Use PSR-7 API instead.
  27     Calling TYPO3\CMS\Extbase\Utility\DebuggerUtility::var_dump() is forbidden, Use logging instead or remove if it was for debugging purposes.
 ------ ---------------------------------------------------------------------------------------------------------------------------------------------
```

oliverklee (approved):

PR #1187, opened 2024-03-04
Migrate to core testing docker image in gitlab images

lukaszuznanski:
For now I removed wait-for-it, I left only `parallel` as additional dependency

https://git.typo3.org/qa/example-extension/-/pipelines/53844

### DIFF
--- a/.gitlab/build/docker_install.sh
+++ b/.gitlab/build/docker_install.sh
@@ -4,4 +4,4 @@
 
 set -xe
 
-apk add parallel
+apk add parallel wait-for-it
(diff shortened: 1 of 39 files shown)


oliverklee:
Could we also use Stefan's image here (to be consistent with `runTests.sh`)? https://github.com/FriendsOfTYPO3/tea/blob/main/Build/Scripts/runTests.sh

lukaszuznanski:
done, https://git.typo3.org/qa/example-extension/-/pipelines/54341

oliverklee (approved):

Issue #1189, opened 2024-03-08
Switch the documentation to PHP-based rendering


oliverklee:
Like this: https://review.typo3.org/c/Packages/TYPO3.CMS/+/83373

PR #1190, opened 2024-03-08
[TASK] Switch the documentation to PHP-based rendering

brotkrueml:
Command for migration:

    docker run --rm --pull always -v $(pwd):/project -it ghcr.io/typo3-documentation/render-guides:latest migrate Documentation

Command for rendering:

    docker run --rm --pull always -v $(pwd):/project -it ghcr.io/typo3-documentation/render-guides:latest --config=Documentation

or just:

    composer docs:generate

The reference to the TYPO3 Documentation Rendering Guide was adjusted as the Quickstart page does not exist anymore. Additionally, the genindex isn't supported anymore. It is planned to consider the `index` directive in the global search on docs.typo3.org.

Resolves: #1189


brotkrueml:
There is currently an issue that `target` on the `image` directive is not considered:
https://github.com/phpDocumentor/guides/issues/911

Therefore, set this PR to draft for the time being.

Documentation/guides.xml
```diff
@@ -0,0 +1,23 @@
...
+    links-are-relative="true"
+>
+    <extension
+        class="\T3Docs\Typo3DocsTheme\DependencyInjection\Typo3DocsThemeExtension"
+        project-home="https://extensions.typo3.org/extension/tea"
+        project-contact="mailto:typo3-coding@oliverklee.de"
+        project-repository="https://github.com/FriendsOfTYPO3/tea"
+        project-issues="https://github.com/FriendsOfTYPO3/tea/issues"
+        edit-on-github-branch="main"
+        edit-on-github="TYPO3-Documentation/tea"
```
oliverklee, line 14: Should this be `FriendsOfTYPO3/tea` instead?

composer.json
```diff
@@ -163,12 +163,7 @@
 		"ci:yaml:lint": "find . ! -path '*.Build/*' ! -path '*node_modules/*' -regextype egrep -regex '.*.ya?ml$' | xargs -r php ./.Build/bin/yaml-lint",
 		"coverage:create-directories": "mkdir -p .Build/logs .Build/coverage",
 		"docs:generate": [
-			"@docs:generate:pullimage",
-			"docker run --rm ghcr.io/t3docs/render-documentation show-shell-commands > tempfile.sh; echo 'dockrun_t3rd makehtml' >> tempfile.sh; bash tempfile.sh; rm tempfile.sh"
-		],
-		"docs:generate:pullimage": [
-			"docker pull ghcr.io/t3docs/render-documentation",
-			"docker tag ghcr.io/t3docs/render-documentation t3docs/render-documentation"
+			"docker run --rm --pull always -v $(pwd):/project -it ghcr.io/typo3-documentation/render-guides:latest --config=Documentation"
```
oliverklee, line 166: If there is only one command, then this doesn't need to be an array anymore.

oliverklee (changes requested):
When running any of the three commands (for migrating, for rendering, and the Composer command for rendering), I get this failure:

```
addgroup: gid '33' in use
adduser: unknown group typo3
su: unknown user typo3
```

Is there anything missing or wrong?

.github/workflows/ci.yml
```diff
@@ -293,3 +293,14 @@ jobs:
           - typo3-version: "^12.4"
             php-version: "8.3"
             composer-dependencies: highest
+  documentation:
+    name: Documentation
+    runs-on: ubuntu-latest
```
oliverklee, line 298: Please use a fixed version here so we won't fail when GitHub updates the `ubuntu-latest` version:

```suggestion
    runs-on: ubuntu-22.04
```

.github/workflows/ci.yml
```diff
@@ -293,3 +293,14 @@ jobs:
...
+    steps:
+      - name: Checkout
+        uses: actions/checkout@v4
+      - name: Test if the documentation will render without warnings
+        run: |
+          mkdir -p Documentation-GENERATED-temp \
+          && docker run --rm --pull always -v $(pwd):/project \
```
oliverklee, line 305: Can we also use the Composer command for this on CI?

.github/workflows/ci.yml
```diff
@@ -293,3 +293,14 @@ jobs:
...
             composer-dependencies: highest
+  documentation:
+    name: Documentation
+    runs-on: ubuntu-latest
+    steps:
+      - name: Checkout
+        uses: actions/checkout@v4
+      - name: Test if the documentation will render without warnings
+        run: |
+          mkdir -p Documentation-GENERATED-temp \
```
oliverklee, line 304: Do we need this `mkdir` here? And if we need it, could/should we integrate it into the Composer command?

oliverklee (changes requested):

oliverklee (changes requested):
We'll also need the corresponding changes to the `docsGenerate` command in `runTests.sh`.

oliverklee:
By the way thanks a lot for stepping up with this change! :heart:

I think this also deserves an entry in the changelog. Could you please add it?

Also, could you please rebase? Thanks!

.github/workflows/ci.yml
brotkrueml: It was necessary some time ago. I think, it is added now to the command. I removed it.

.github/workflows/ci.yml
brotkrueml: Done

.github/workflows/ci.yml
```diff
@@ -301,5 +301,4 @@ jobs:
         uses: actions/checkout@v4
       - name: Test if the documentation will render without warnings
         run: |
-          && docker run --rm --pull always -v $(pwd):/project \
-             ghcr.io/typo3-documentation/render-guides:latest --config=Documentation --no-progress --fail-on-log
+          composer docs:generate
```
oliverklee, line 304: In cases where there's only one line for `run`, we can also drop the `|` notation and write the command directly after `run:`.

brotkrueml:
> We'll also need the corresponding changes to the `docsGenerate` command in `runTests.sh`.

Currently, I don't understand what is going on here (why writing a new shell script)? Is it sufficient to just add the new command?

.github/workflows/ci.yml
brotkrueml: Done, personally, I like it with the "|" as it is better readable IMHO.

oliverklee: I'm fine either way (even if I prefer the one-line version). So please go ahead and use the version that you like best. :-)

oliverklee:
> Currently, I don't understand what is going on here (why writing a new shell script)?

The idea of `runTests.sh` is that people can develop the extension and run all relevant tasks (run unit tests, run PHPStan, generate documentation …) without needing to have a local PHP and Composer installation (including having the right PHP extensions installed), and that they can use either Docker or Podman.

brotkrueml:
> > Currently, I don't understand what is going on here (why writing a new shell script)?
> 
> The idea of `runTests.sh` is that people can develop the extension and run all relevant tasks (run unit tests, run PHPStan, generate documentation …) without needing to have a local PHP and Composer installation (including having the right PHP extensions installed), and that they can use either Docker or Podman.

Yup, I know runTests.sh, but not what is going on in the docsGenerate command. Why this strange "we write it to a shell script and execute this then" instead of just running the original command? Additionally, the documentation rendering (old and new) are running in a container, so there is already no need to have PHP and stuff installed locally - and facade it behing the runTests.sh script.

oliverklee:
A note on updating PRs: Please always **rebase** on `main` instead of merging the latest `main` into the feature branch. Merging `main` into feature branches creates hard-to-resolve problems when there are merge conflicts. (Unfortunately, I don't know of a way to disable to GUI features for merging `main` into PR branches in GitHub.)

.github/workflows/ci.yml
brotkrueml: The composer command won't work, as it uses an interactive shell, which GitHub Actions does not like. Therefore the original command was set in place again.

oliverklee:
> Yup, I know runTests.sh, but not what is going on in the docsGenerate command.

I don't either. ;-) It's okay for me if you ditch the current (Docker-specific) part of that command in `runTests.sh` and replace it with something working.

oliverklee:
@brotkrueml Any news on this PR? I'd really like to get it in.

brotkrueml:
What is the reason to use the docs command in runTests.sh anyway? The runTests.sh is about preparing and running the PHP tests.
As docker can't run in docker, maybe that's why that thing with the "write to a shell script" is there.
I would like to remove it from runTests.sh as we provide our own GitHub action for checking for warning on rendering.

oliverklee:
> What is the reason to use the docs command in runTests.sh anyway? 

I'm not completely sure - it may have been a "copy paste" thing, a "implement all Composer scripts" or something historic.

What I'd like to have is a single, easy-to-remember, no-need-for-copy'n'pasting command to run on my machine to build the docs even if I don't have a local Composer of PHP installed. (I personally have PHP and Composer on my machine, but we don't want to require developers to have that, particularly if they need a particular PHP version that may not be available on Ubuntu LTS systems.)

brotkrueml:
Sorry, can't help with the docker thingy. In the Documentation Team we use a Makefile and run "make docs" (if this helps).

DanielSiepmann:
I've rebased and adopted the `runTest.sh` entry.

oliverklee (approved):

Issue #1186, opened 2024-03-04
Put all tooling configuration files into one directory .config


lukaszuznanski:
We should put all configuration files of tooling, which includes testing, to one directory which is called `config`

PR #1202, opened 2024-03-25
Move all tooling configuration to Build dir

lukaszuznanski:
Resolve: #1186


oliverklee:
I think this deserves an entry in the changelog. What do you think?

oliverklee:
@sbuerk provided some more feedback and recommends to __not__ use `config/` for this as this by convention is the production configuration folder for TYPO3 sites. I'll put this up for decision in the next call.

lukaszuznanski:
Ok, I'm not sure if this was decided, can't find it in meeting minutes

lukaszuznanski:
@oliverklee @sbuerk should we use `Build/` ?

sbuerk:
> @oliverklee @sbuerk should we use `Build/` ?

If you ask me, yes. That's also the folder the TYPO3 core uses for these things of tool configurations (with sub-folders per tool) and core merger/contributors adopted for extensions and projects.

I talked with @oliverklee and @DanielSiepmann on that already. In the end, it's about naming with some requirements:

* folder to keep root-folder clean
* not using a folder name which contains extension related stuff (ex. `Configuration/` ) in the TYPO3 world

There are other preferences, which you should discuss/define within the team first I guess. For example stying on the root-folder / default naming.

Personally, `Build/` suits most perfectly for these kind of stuff. I would stay way from things like `etc/` or namings used in other contexts not suiting the need.

oliverklee:
I propose we split this PR up into "one tool per PR". This makes the reviews easier, reduces the risk of us missing things, and helps get the changes merged faster.

PR #1207, opened 2024-04-01
[TASK] Use the development PHP INI on CI

### LINKED ISSUE #1197: Use the development PHP INI on CI
This will allow us to see (and error out on) all warnings and notices.
https://github.com/shivammathur/setup-php/issues/827

oliverklee:
This allows us to see (and fail on) more PHP warnings and notices.

Fixes #1197

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -22,6 +22,7 @@ jobs:
         uses: shivammathur/setup-php@v2
         with:
           php-version: "${{ matrix.php-version }}"
+          ini-file: development
           coverage: none
           tools: composer:v2
       - name: "Show the Composer configuration"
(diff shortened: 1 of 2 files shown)


oliverklee:
Haha, this has revealed that we'll need a newer version of  `webmozart/assert` to avoid the warning that was hidden before. Marking as draft for now. I'll create a PR for the `webmozart/assert` requirement.

oliverklee:
Using the newer `webmozard/assert` solves most of the problems. We still have one more failure we'll need to solve.

lukaszuznanski (commented):
From my side, it's OK, but there is gitlab part missing, so let's merge this one and I will create separate issue for gitlab.

DanielSiepmann (approved):

PR #1212, opened 2024-04-02
[BUGFIX] Raise the minimal TYPO3 V12 version

oliverklee:
TYPO3 12.4.2 has some bug fixes for PHP 8.3 to avoid a class loading issue.


oliverklee:
Note for testing: This is what got #1207 to work.

DanielSiepmann (approved):

PR #1213, opened 2024-04-02
Drop support for predefined CI workflow

lukaszuznanski:
Resolves: #1210


oliverklee:
Sorry, I took the ticket and forgot to ping you. So we now have another PR for this that also updates the docs: #1211

PR #1223, opened 2024-04-14
[TASK][!!!] Require a storage PID for the tea list

### LINKED ISSUE #1024: Make the repository use the storage PID

oliverklee:
Fixes #1024
Supersedes #1024


oliverklee:
I am currently creating a bunch of pre-patches for this.

Tests/Functional/Controller/Fixtures/TypoScript/Constants/PluginConfiguration.typoscript
```diff
@@ -0,0 +1,6 @@
+plugin.tx_tea {
+  persistence {
+    # cat=plugin.tx_tea//a; type=string; label=Default storage PID
```
DanielSiepmann, line 3: I guess the format is coming from ext configuration files and isn't needed here within the fixture?

oliverklee: Done.

DanielSiepmann (approved):

PR #1232, opened 2024-04-15
[TASK] Add `DebugUtility::debug()` to list of `disallowedStaticCalls`

### LINKED ISSUE #1231: Add \TYPO3\CMS\Core\Utility\DebugUtility::debug to disallowedStaticCalls
Hi,
please add the static  method \TYPO3\CMS\Core\Utility\DebugUtility::debug() to the list of disallowedStaticCalls in phpstan.neon

If you like, I can prepare a PR.

cschnell:
Fixes #1231


phpstan.neon
```diff
@@ -46,7 +46,9 @@ parameters:
       message: 'Use PSR-7 API instead'
   disallowedStaticCalls:
     -
-      method: 'TYPO3\CMS\Extbase\Utility\DebuggerUtility::var_dump()'
+      method:
+        - 'TYPO3\CMS\Extbase\Utility\DebuggerUtility::var_dump()'
+        - '\TYPO3\CMS\Core\Utility\DebugUtility::debug()'
```
oliverklee, line 51: This should be consistent - either with or without a leading backslash.

oliverklee (changes requested):

phpstan.neon
cschnell: Absolutely. I have added the backslash to the DebuggerUtility, because it is the FQDN.

oliverklee: I've just had a look into the documentation of the PHPStan extension: There, they use it without the trailing backslash. I'd propose we stick to that.

https://github.com/spaze/phpstan-disallowed-calls/blob/main/docs/custom-rules.md

(I've also just tested that this works.)

oliverklee (changes requested):

phpstan.neon
cschnell: Right, done.

oliverklee (approved):
Looks good. Thanks for contributing, @cschnell! :heart:

PR #1242, opened 2024-04-17
[CLEANUP] Reduce redundancies in the PHPStan configuration

oliverklee:
Directories configured via `paths` already allows PHPStan to discover the symbols in the code located there. So there is no need to add these directories again via `scanDirectories`.

https://phpstan.org/user-guide/discovering-symbols#third-party-code-outside-of-composer-dependencies

### DIFF
--- a/phpstan.neon
+++ b/phpstan.neon
@@ -20,11 +20,6 @@ parameters:
     - Configuration
     - Tests
 
-  scanDirectories:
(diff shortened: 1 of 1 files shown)


oliverklee:
Thanks, I've entered #1245 concerning `bootstrapFiles`.

Issue #1252, opened 2024-04-28
Raise PHPStan to level 9


oliverklee:
For some reason, we're not on level 9 yet. Let's raise the level one by one and fix any issues along the way.

DanielSiepmann:
I guess you want level 9 instead of max as that gives us more control during updates of PHPStan?
We actually always use max level in our projects and extensions.

oliverklee:
> I guess you want level 9 instead of max as that gives us more control during updates of PHPStan?

Exactly, that is my thinking behind this. (But I'd also be fine with using max level.)

DanielSiepmann:
I'm fine with level 9. Just one note: It might happen that PHPStan adds level 10, 11, etc. So we probably should have an eye on this.

PR #1258, opened 2024-04-30
[TASK] Raise PHPStan to level 6

oliverklee:
Part of #1252

### DIFF
--- a/phpstan-baseline.neon
+++ b/phpstan-baseline.neon
@@ -1,2 +1,6 @@
 parameters:
-	ignoreErrors: []
+	ignoreErrors:
+		-
+			message: "#^Method TTN\\\\Tea\\\\Domain\\\\Repository\\\\TeaRepository\\:\\:findByOwnerUid\\(\\) return type with generic interface TYPO3\\\\CMS\\\\Extbase\\\\Persistence\\\\QueryResultInterface does not specify its types\\: ModelType$#"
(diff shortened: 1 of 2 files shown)


phpstan-baseline.neon
```diff
@@ -1,2 +1,6 @@
 parameters:
-	ignoreErrors: []
+	ignoreErrors:
+		-
+			message: "#^Method TTN\\\\Tea\\\\Domain\\\\Repository\\\\TeaRepository\\:\\:findByOwnerUid\\(\\) return type with generic interface TYPO3\\\\CMS\\\\Extbase\\\\Persistence\\\\QueryResultInterface does not specify its types\\: ModelType$#"
```
DanielSiepmann, line 4: I don't understand why we need this, can you explain?
I'd expect saschaegerer/phpstan-typo3 to take care of this. Furthermore we annotated the repository: https://github.com/FriendsOfTYPO3/tea/blob/main/Classes/Domain/Repository/TeaRepository.php#L13

oliverklee: We probably need to add a return type annotation using the generic for this method. (As far as I understand, saschaegerer/phpstan-typo3 only works for the methods that the repository base class already has.)

oliverklee:
Ping

DanielSiepmann (approved):

PR #1270, opened 2024-05-06
[TASK] Narrow down a type in the controller

oliverklee:
This fixes the last PHPStan level 9 warning.

### DIFF
--- a/Classes/Controller/FrontEndEditorController.php
+++ b/Classes/Controller/FrontEndEditorController.php
@@ -42,7 +42,10 @@ class FrontEndEditorController extends ActionController
      */
     private function getUidOfLoggedInUser(): int
     {
-        return $this->context->getPropertyFromAspect('frontend.user', 'id');
+        $userUid = $this->context->getPropertyFromAspect('frontend.user', 'id');
+        \assert(\is_int($userUid) && $userUid >= 0);
+
+        return $userUid;
     }
 
     /**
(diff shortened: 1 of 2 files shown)


Classes/Controller/FrontEndEditorController.php
```diff
@@ -42,7 +42,10 @@ public function indexAction(): ResponseInterface
      */
     private function getUidOfLoggedInUser(): int
     {
-        return $this->context->getPropertyFromAspect('frontend.user', 'id');
+        $userUid = $this->context->getPropertyFromAspect('frontend.user', 'id');
+        \assert(\is_int($userUid) && $userUid >= 0);
```
DanielSiepmann, line 46: I'd be happy if we could define best practices to handle such things.
E.g. whether to use native assert (with recommendation on PHP configuration) or whether to use some assert library.

I'm curious myself and didn't find a clear answer for myself yet.

DanielSiepmann (approved):

Classes/Controller/FrontEndEditorController.php
oliverklee: Yes, let's please do this. Would you be willing to put this on the agenda of our next team call?

DanielSiepmann: Done: https://notes.typo3.org/JKT4Cr6tSZehjq0Nu8o65g?both#discussion-checking-type-and-value-of-variables-assertions-exceptions-daniel

Issue #1280, opened 2024-05-07
Cannot connect to the MySQL server in `runTests.sh` with Docker and with MySQL 8.0


oliverklee:
```bash
./Build/Scripts/runTests.sh -s functional -b docker -d mysql -i 8.0
```

```
Using driver: mysqli
Can not connect to mysql-func-32625 port 3306. Aborting.
PHPUnit 9.6.19 by Sebastian Bergmann and contributors.

EEEEEEEEEEEEE                                                     13 / 13 (100%)

Time: 00:01.010, Memory: 87.00 MB

There were 13 errors:

1) TTN\Tea\Tests\Functional\Controller\TeaControllerTest::indexActionRendersAllAvailableTeas
Doctrine\DBAL\Exception\ConnectionException: An exception occurred in the driver: Connection refused

/home/klee/src/typo3/ext/tea/.Build/vendor/doctrine/dbal/src/Driver/API/MySQL/ExceptionConverter.php:101
/home/klee/src/typo3/ext/tea/.Build/vendor/doctrine/dbal/src/Connection.php:1939
/home/klee/src/typo3/ext/tea/.Build/vendor/doctrine/dbal/src/Connection.php:1887
/home/klee/src/typo3/ext/tea/.Build/vendor/doctrine/dbal/src/Connection.php:380
/home/klee/src/typo3/ext/tea/.Build/vendor/typo3/cms-core/Classes/Database/Connection.php:98
/home/klee/src/typo3/ext/tea/.Build/vendor/doctrine/dbal/src/Connection.php:453
/home/klee/src/typo3/ext/tea/.Build/vendor/doctrine/dbal/src/Connection.php:411
/home/klee/src/typo3/ext/tea/.Build/vendor/doctrine/dbal/src/Connection.php:318
/home/klee/src/typo3/ext/tea/.Build/vendor/typo3/cms-core/Classes/Database/ConnectionPool.php:233
/home/klee/src/typo3/ext/tea/.Build/vendor/typo3/cms-core/Classes/Database/ConnectionPool.php:173
/home/klee/src/typo3/ext/tea/.Build/vendor/typo3/testing-framework/Classes/Core/Testbase.php:631
/home/klee/src/typo3/ext/tea/.Build/vendor/typo3/testing-framework/Classes/Core/Functional/FunctionalTestCase.php:408
/home/klee/src/typo3/ext/tea/Tests/Functional/Controller/TeaControllerTest.php:25

Caused by
[...]

PR #1285, opened 2024-05-07
[BUGFIX] Ensure correct service waits and container cleanup

sbuerk:
This change now uses a bash trap to ensure proper container
cleanup after some signals happend. The internal `waitFor()`
is modified to emit a signal to cleanup containers if the
waiting process does not process correctly and stop the
script execution.

Additionally, `--pull-never` as CI_PARAMS is no longer derived
from the `$CI` variable and must now be handed over from the
calling process.

Further, the max retry value in `waitFor()` is increased to 20
to mitigate the slowed down startup of MySQL 8.0 container.

Resolves: #1280

### DIFF
--- a/Build/Scripts/runTests.sh
+++ b/Build/Scripts/runTests.sh
@@ -4,13 +4,15 @@
 # TYPO3 core test runner based on docker.
 #
 
+trap 'cleanUp;exit 2' SIGINT
+
 waitFor() {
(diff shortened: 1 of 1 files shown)


oliverklee (changes requested):
Now this works. Thanks!

oliverklee (approved):

Issue #1172, opened 2024-02-19
Migrate Services.yaml to Services.php


lukaszuznanski:
Following best practices, migrate from .yaml to .php file

DanielSiepmann:
I'd wait for https://notes.typo3.org/JKT4Cr6tSZehjq0Nu8o65g?both#decision-and-to-dos-separate-concepts-from-implementation-details-daniel before migrating. We should document that DI is best practice and why we use `Services.php` over `Services.yaml`. What are pros- and cons of both ways.

Issue #1237, opened 2024-04-15
Add information to README that we want to align with core


lukaszuznanski:
integrate information into READMe that we want to technically align with core

PR #1286, opened 2024-05-07
[TASK] Migrate from Services.yaml to Services.php

DanielSiepmann:
* Migrate existing configuration from yaml to php.
* Document why we did the switch and that using yaml or both is totally valid and conforms to best practices.
* Document why we are not using attributes but only `Services.php` right now.

Resolves: #1172
Relates: #1237

### DIFF
--- a/Configuration/Services.php
+++ b/Configuration/Services.php
@@ -1,0 +1,15 @@
+<?php
+
+declare(strict_types=1);
+
+namespace Symfony\Component\DependencyInjection\Loader\Configurator;
+
+return static function (ContainerConfigurator $containerConfigurator) {
+    $services = $containerConfigurator->services()
+        ->defaults()
+        ->autowire()
+        ->autoconfigure();
+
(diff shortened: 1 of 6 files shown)


sbuerk (approved):

Configuration/Services.php
```diff
@@ -0,0 +1,15 @@
...
+namespace Symfony\Component\DependencyInjection\Loader\Configurator;
+
+return static function (ContainerConfigurator $containerConfigurator) {
+    $services = $containerConfigurator->services()
+        ->defaults()
+        ->autowire()
+        ->autoconfigure()
+    ;
+
+    $services->load('TTN\\Tea\\', '../Classes/*');
```
brotkrueml, line 14: The exclude is missing from Services.yaml

DanielSiepmann: Thanks, good catch. I've added the exclude.

PR #1287, opened 2024-05-07
[TASK] Use PHP 8.3 by default in `runTests.sh`

oliverklee:
Let's use the most performant available PHP version for local development.

### DIFF
--- a/Build/Scripts/runTests.sh
+++ b/Build/Scripts/runTests.sh
@@ -223,9 +223,9 @@ Options:
         Specifies the PHP minor version to be used
             - 7.4: use PHP 7.4
             - 8.0: use PHP 8.0
-            - 8.1: (default) use PHP 8.1
+            - 8.1: use PHP 8.1
             - 8.2: use PHP 8.2
-            - 8.3: use PHP 8.3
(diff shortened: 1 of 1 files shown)


sbuerk (approved):
Personally I prefere to use the minimum version locally to ensure early (locally) not to use syntax only available in higher versions.
Performance in runTests.sh for tests and tools should be negitable.

On the otherhand, it's a question of preference and technically absolutly fine to use PHP8.3 as both cores support them and all related tools.

lukaszuznanski (approved):
I can confirm that it's working on local env.

DanielSiepmann:
If this is about performance: Should we do the same for all other versions, e.g. database as well?
mysql for some reason is the only dbms where the highest version is the default.

I'd prefer to streamline the highest / lowest version throughout all the usages. TYPO3 has highest value by default. Or should that be a different issue?

oliverklee:
> I'd prefer to streamline the highest / lowest version throughout all the usages. TYPO3 has highest value by default. Or should that be a different issue?

Let's do that for the DBMS versions in a separate issue. (And I'm all for using the highest possible version for the supported TYPO3 versions by default.)

Issue #1235, opened 2024-04-15
Create issue templates for providing feedback


lukaszuznanski:
Create issue templates for providing technical feedback on things that people don't agree with.

PR #1291, opened 2024-05-07
[TASK] Add "Technical Feedback" issue template

DanielSiepmann:
Allow others to easily provide feedback on technical decisions. The issue template should make people more invited to share their own feedback.

Resolves: #1235


sbuerk:
Late +1 ;)

Issue #1120, opened 2024-01-16
Add a command to create test data


oliverklee:
## Goal
I as a user would like to create test records.

## Acceptance Criteria

- [ ] a command controller is added for creating the test records in a existing sysfolder. Pid of this folder as option
- [ ] the command controller has an option to delete existing records in this sysfolder.

## Additional Information
For manual testing, it would be helpful to have a (Symfony?) command to create some test records.

This could also server as an example for creating commands.

The script should also throw away old test data.

kanow:
my idea:

- create a command controller to creating new data in an existing sysfolder. pid must be set as argument
- add second option to delete existing data in that pid
- maybe later extend the command or use another to create that syfolder before. But maybe that's not useful. I'm not sure.

oliverklee:
Sounds generally good to me.

Also, I think we can always delete existing data. This is only for setting up test data quickly and not for usage on a production site.

PR #1297, opened 2024-05-07
[FEATURE] Add command controller to create test data

kanow:
Simple command to add test data for tea model with a console command. It's just the first try and should be extended by more data, translations, etc.

Resolves: #1120

### DIFF
--- a/Classes/Command/CreateTestDataCommand.php
+++ b/Classes/Command/CreateTestDataCommand.php
@@ -34,6 +34,7 @@ final class CreateTestDataCommand extends Command
             'sys_language_uid' => 0,
         ],
     ];
+
     protected function configure(): void
     {
         $this
--- a/Classes/Command/CreateTestDataCommand.php
+++ b/Classes/Command/CreateTestDataCommand.php
@@ -4,13 +4,19 @@ declare(strict_types=1);
 
 namespace TTN\Tea\Command;
 
+use Symfony\Component\Console\Application;
 use Symfony\Component\Console\Command\Command;
 use Symfony\Component\Console\Input\InputArgument;
 use Symfony\Component\Console\Input\InputInterface;
 use Symfony\Component\Console\Input\InputOption;
 use Symfony\Component\Console\Output\OutputInterface;
+use Symfony\Component\Console\Style\SymfonyStyle;
+use TYPO3\CMS\Backend\Command\ProgressListener\ReferenceIndexProgressListener;
+use TYPO3\CMS\Backend\Command\ReferenceIndexUpdateCommand;
+use TYPO3\CMS\Core\Core\Bootstrap;
 use TYPO3\CMS\Core\Database\Connection;
 use TYPO3\CMS\Core\Database\ConnectionPool;
+use TYPO3\CMS\Core\Database\ReferenceIndex;
 use TYPO3\CMS\Core\Utility\GeneralUtility;
 
 /*
@@ -75,6 +81,10 @@ final class CreateTestDataCommand extends Command
             );
         }
         $output->writeln(sprintf('Test data in page %s created.', $pageUid));
+
+        $referenceIndex = GeneralUtility::makeInstance(ReferenceIndex::class);
+        $referenceIndex->updateIndex(0);
+        $output->writeln('Reference index updated.');
 
         return Command::SUCCESS;
     }
--- a/Classes/Command/CreateTestDataCommand.php
+++ b/Classes/Command/CreateTestDataCommand.php
@@ -4,16 +4,11 @@ declare(strict_types=1);
 
 namespace TTN\Tea\Command;
 
-use Symfony\Component\Console\Application;
 use Symfony\Component\Console\Command\Command;
 use Symfony\Component\Console\Input\InputArgument;
 use Symfony\Component\Console\Input\InputInterface;
 use Symfony\Component\Console\Input\InputOption;
 use Symfony\Component\Console\Output\OutputInterface;
-use Symfony\Component\Console\Style\SymfonyStyle;
-use TYPO3\CMS\Backend\Command\ProgressListener\ReferenceIndexProgressListener;
-use TYPO3\CMS\Backend\Command\ReferenceIndexUpdateCommand;
-use TYPO3\CMS\Core\Core\Bootstrap;
 use TYPO3\CMS\Core\Database\Connection;
 use TYPO3\CMS\Core\Database\ConnectionPool;
 use TYPO3\CMS\Core\Database\ReferenceIndex;
--- a/Classes/Command/CreateTestDataCommand.php
+++ b/Classes/Command/CreateTestDataCommand.php
@@ -19,7 +19,9 @@ use TYPO3\CMS\Core\Utility\GeneralUtility;
 
 final class CreateTestDataCommand extends Command
 {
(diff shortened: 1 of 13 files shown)


sbuerk:
I know this is WIP, but I wanted to provide an early feedback on this.

Basically, I would recommend to avoid using `QueryBuilder` or `Connection`
methods (low-level api) for such tasks, until absolutly needed. Usually,
it's more recommandable to use `DataHandler` for operation, as it takes
care of a lot of stuff. For example:

* Default value handling (TCA)
* TYPO3 system field handling
* soft-delete/hard-delete of records based on TCA configuration
* record history for changes (soft-delete, creation, updating, ...)
* automatically exeucte hooks configured by extensions/core
* workspace support if workspace is installed etc without the need to handle system fields directly
* system field for localization of records (l18n_diff, l18n_source) etc
* access verification (if not executed with admin permissions)

albeit eventually not having all in place right now.

Thus I highly recommend to switch to `DataHandler` for the deletion
and data creation/update part.

Classes/Command/CreateTestDataCommand.php
```diff
@@ -0,0 +1,77 @@
...
+
+declare(strict_types=1);
+
+namespace TTN\Tea\Command;
+
+/*
+ * Command to create test data for the tea extension.
+ */
+
+use phpDocumentor\Reflection\Types\Integer;
```
sbuerk, line 11: I assume that this is wrong - what's the reason to use the phpDocumentor reflection type here ? Maybe some error with auto import ?

kanow: Yes, now removed.

kanow:
We are aware to the usage of DataHandler or not. Daniel means that Command Controller should be an simple example to work with database actions. Maybe later we can extend that existing command or create a second command controller that shows an advanced example to work with the DataHandler.
For example adding more data with relations to images.

Classes/Command/CreateTestDataCommand.php
```diff
@@ -0,0 +1,79 @@
+<?php
+
+declare(strict_types=1);
+
+namespace TTN\Tea\Command;
+
+/*
+ * Command to create test data for the tea extension.
+ */
```
oliverklee, line 9: This class comment should go directly above the class.

Classes/Command/CreateTestDataCommand.php
```diff
@@ -0,0 +1,79 @@
...
+use Symfony\Component\Console\Input\InputInterface;
+use Symfony\Component\Console\Input\InputOption;
+use Symfony\Component\Console\Output\OutputInterface;
+use TYPO3\CMS\Core\Database\Connection;
+use TYPO3\CMS\Core\Database\ConnectionPool;
+use TYPO3\CMS\Core\Utility\GeneralUtility;
+
+final class CreateTestDataCommand extends Command
+{
+    /** @phpstan-ignore-next-line */
```
oliverklee, line 22:
```suggestion
    /**
     * @var list<array{title: non-empty-string, description: non-empty-string, sys_language_uid: int<0, max>}>
     */
```

Classes/Command/CreateTestDataCommand.php
```diff
@@ -0,0 +1,79 @@
...
+            'description' => 'I love that tea!',
+            'sys_language_uid' => 0,
+        ],
+         [
+            'title' => 'Earl Grey',
+            'description' => 'A nice tea!',
+            'sys_language_uid' => 0,
+        ],
+    ];
+    protected function configure(): void
```
oliverklee, line 35:
```suggestion

    protected function configure(): void
```

Classes/Command/CreateTestDataCommand.php
```diff
@@ -0,0 +1,79 @@
...
+    {
+        $this
+            ->setHelp('Create test data for the tea extension in an already existing page (sysfolder).')
+            ->addArgument(
+                'pageId',
```
oliverklee, line 40: This probably should be `pageUid`, not `pageId`.

Classes/Command/CreateTestDataCommand.php
```diff
@@ -0,0 +1,79 @@
...
+                'delete-data-before',
+                'd',
+                InputOption::VALUE_NONE,
+                'Delete all tea data in the defined pid before creating new data.'
+            );
+    }
+
+    protected function execute(InputInterface $input, OutputInterface $output): int
+    {
+        /** @var int $pageId */
```
oliverklee, line 54: In production code, please let's use `assert` to state data restrictions that we (as human) know to be true.

Classes/Command/CreateTestDataCommand.php
```diff
@@ -0,0 +1,79 @@
...
+        $pageId = $input->getArgument('pageId') ?? 0;
```
oliverklee, line 55: `pageUid`

Classes/Command/CreateTestDataCommand.php
```diff
@@ -0,0 +1,79 @@
...
+        /** @var bool $deleteDataBefore */
```
oliverklee, line 56: Ditto about `assert`.

Classes/Command/CreateTestDataCommand.php
```diff
@@ -0,0 +1,79 @@
...
+        $deleteDataBefore = $input->getOption('delete-data-before') ?? false;
+        $table = 'tx_tea_domain_model_tea';
+        $connectionForTable = GeneralUtility::makeInstance(ConnectionPool::class)->getConnectionForTable($table);
+
+        if ($deleteDataBefore) {
+            $query = $connectionForTable;
+            $query->delete($table, ['pid' => $pageId], [Connection::PARAM_INT]);
+            $output->writeln(sprintf('Existing data in page %s deleted.', $pageId));
```
oliverklee, line 64: Please let's use the `DataHandler` for all delete and insert operations so indices and cache clearing are used. Feel free to steal from here: https://github.com/oliverklee/ext-seminars/blob/main/Classes/Domain/Repository/Event/EventRepository.php#L291

Classes/Command/CreateTestDataCommand.php
```diff
@@ -0,0 +1,79 @@
...
+use Symfony\Component\Console\Command\Command;
+use Symfony\Component\Console\Input\InputArgument;
+use Symfony\Component\Console\Input\InputInterface;
+use Symfony\Component\Console\Input\InputOption;
+use Symfony\Component\Console\Output\OutputInterface;
+use TYPO3\CMS\Core\Database\Connection;
+use TYPO3\CMS\Core\Database\ConnectionPool;
+use TYPO3\CMS\Core\Utility\GeneralUtility;
+
+final class CreateTestDataCommand extends Command
```
oliverklee, line 20: Please let's have functional tests for this command.

oliverklee (changes requested):

oliverklee (changes requested):
Currently, the functional tests are failing. Also, we currently have some style warnings and PHPStan warnings which we'll need to fix first.

oliverklee (changes requested):
Currently, some tests are red. Could you please fix these and also rename the Composer script key as requested in an earlier review? Thanks! :pray:

kanow:
Oh, shit. And yes, I still try to make this PR ready to review/merge ;-).

oliverklee (approved):

Issue #1301, opened 2024-05-07
Technical Feedback: NODEJS/NPM dependency managemant requirements


sbuerk:
**Describe the area**

In the past, `yarn` has been used as dependency management tool
for the `nodejs` package and tool chain ecosystem. yarn requires
binary blob handling for major version 2 and higher, which can't
be directly installed - makeing that handling quite anoying.

Meanwhile, the ecosystem has envolved and the native dependency
management tool `npm` envolved along with a proper `package-lock.json`
file management.

It's recommended to use either yarn or npm and do not mix them.

Using the native toolchain ...

1. ... it is required to ensure at least `npm 9.x` version along with
   a suitable nodejs version. 
2. ... the `package-lock.json` file should be commited to ensure correct usage
   and no auto updating to avoid inconsistency in builds (pr / merges).

**Feedback**

With #1289 the `yarn` dependency manager for nodejs libraries
and tools has been replaced with direct `npm` usage. This is
generally a good move, but the change missed some important
things:

* `nodejs version` is not enforced
* `npm version` is not enforced
* `package-lock.json` file is not commited in the repository,
  which should be done to ensure reproducable installs
  in changes/schedules.
  
To ensure nodejs and npm version, a `engine` block can be
added to the `package.json` - which would npm/nodejs fail
if used with the wrong version. For example:

```json
{
  "engines": {
    "node": ">=18.19.0 <19.0.0",
    "npm": ">=10.0.0"
  },  
}
```

Limiting the nodejs/npm version also that local builds are
not done with unsupported versions (which may install other
dependencies of no package.lock file is used). Limiting also
allows to provide a concrete check in runTests.sh and CI
actions.

For local development, until runTests.sh support for npm and
js/css builds has been added, local installations needs to be
used. To help local management with multiple nodejs/npm versions,
the `nvm` (node version manager) can be used - which comes with
`nvmrc` file support to select/use correct nodejs/version in a
switchable manner if some shell (bash) handling scripts is added.
That means, if nvm-autochange shell extension is setup, the npm
and nodejs version is automatically switched to the defined version
in a folder/subfolder.

`npm ci` can be used to install exactly the same version from the
lock-file (but requires a existing one) and thus forces to maintain
a commited lock file (which is a good point).

A tricky point is, that there are different `package-lock.json` formats
out there and at least version 3 is required to have a correct dependency
management making `yarn` obsolete. This can be ensured with a `.npmrc`
file.

[...]

PR #1289, opened 2024-05-07
Switch from yarn to npm

lukaszuznanski:
Resolves: #1274

PR #1302, opened 2024-05-07
[TASK] Streamline `nodejs` dependency management

sbuerk:
With #1289 the `nodejs` dependency management has
been changed from `yarn` to `npm` missing the one
or other important requirement.

This change streamlines the management by ...

* adding a `.nvmrc` file to the repository root to
  allow automatic nodejs/npm switch if the nvm shell
  switching is available on the host system.
* adding a `.npmrc` file to specify the lock file
  version and engine option.
* adding `nodejs` and `npm` version constraints as
  `engine` specification to the `package.json` file.
* remove `package-lock.json` from `.gitignore` and
  add it to the repository to ensure reproducable
  setups - which becomes more important if extension
  get custom backend/frontend modules and javascript.
* adding `package-lock.json` to exclude it from archive,
  due to remove from the `.gitignore` file.
* ensuring that new development files are excluded from
  packaging and publishing.    
* use `npm ci` in GitHub action workflows to install
  from the lock-file.

Note: This change unblocks adding `npm` dispatching
to `Build/Scripts/runTests.sh`.

Resolves: #1301
Related: #1289


package.json
```diff
@@ -9,6 +9,10 @@
 	"author": "",
 	"version": "1.0.0",
 	"license": "ISC",
+	"engines": {
+		"node": ">=18.19.0 <19.0.0",
```
lukaszuznanski, line 13: is there a reason to lock it to ```<19.0.0``` ? I was testing it using ```21.x```.
Should .nvmrc be enough for setting node version? Then we can maybe widen this requirement a bit.

sbuerk: If you use nodejs 22 for example to add dependency, it could happen that a version is installed/added to the `package.json` / `package-lock.json` not installable with an
older nodejs/npm version.

The nodejs/npm ecosystem is quit sensible in this area. Thus we have it nailed down to specific nodejs/npm version in the core to avoid such issues here.

And it allows a smaller diverging chain across `.nvmrc` and eventually docker image using (later in runTest.sh` as it enforces to keep them all in the same line.

On the otherhand - I'm not a nodejs ecosystem expert. So no hard opionen on this in general beside opening for missmatch between local and ci runs.

.npmrc
```diff
@@ -0,0 +1,2 @@
+lockfile-version=3
```
oliverklee, line 1: Should this files also go into the `.gitattributes` to keep it out of packaging? Or should we package it?

.gitattributes
```diff
@@ -14,6 +14,7 @@
 /Tests/ export-ignore
 /eslint.config.json export-ignore
 /package.json export-ignore
+/package-lock.json export-ignore
```
oliverklee, line 17: We also need to add this file to the list of files in the `prepare-release` Composer script (to keep it out of the TER when someone does a manual release).

.npmrc
sbuerk: ack. will do so.

.gitattributes
sbuerk: ack. will do so.

sbuerk: done.

.npmrc
sbuerk: done.

oliverklee (approved):
Looks good to me. Thanks!

(I'd like to have @lukaszuznanski have another look as well.)

lukaszuznanski (approved):

PR #1339, opened 2024-06-10
[TASK] Document our approach to assertions and type checks

### LINKED ISSUE #1330: Document checking type and value of variables: assertions, exceptions
Document why we should use assert() instead of dedicated libraries

oliverklee:
Fixes #1330

### DIFF
--- a/Documentation/Assertions.rst
+++ b/Documentation/Assertions.rst
@@ -1,0 +1,72 @@
+.. include:: /Includes.rst.txt
+
+.. _assertions-and-type-checks:
+
+==========================================
+Our approach to assertions and type checks
+==========================================
+
+In general, we use assertions and type checks to catch bad input, to help debug
+problems, and to help static type analysis tools like PHPStan or Psalm to
+determine the type and range of a variable where it cannot detect it
+automatically.
+
+In production code
+------------------
+
+To catch bad input that actually is possible, we throw an exception. This allows
+developers and users to see what is wrong and to fix it.
+
+..  code-block:: php
+
+    if (!$model instanceof Tea) {
+        throw new \RuntimeException('No model found.', 1687363745);
+    }
+
+
+To help PHPStan and Psalm with variables where we (as developers) know the type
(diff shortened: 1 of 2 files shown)


lukaszuznanski:
Should this way for decision for ADR/Spikes and then document it this way or documentation is just fine?

oliverklee:
In our [last call](https://git.typo3.org/qa/internal-organization/-/blob/master/MeetingMinutes/2024/05/27.md?ref_type=heads), we decided this:

> Oli already has experience with that and will add a PR to EXT:tea documentation to document the best practices, so we all can learn and follow the advice.

So we can use this as-is without requiring an additional decision.

Documentation/Assertions.rst
```diff
@@ -0,0 +1,72 @@
+.. include:: /Includes.rst.txt
+
+.. _assertions-and-type-checks:
+
+==========================================
+Our approach to assertions and type checks
+==========================================
```
DanielSiepmann, line 7:
```suggestion
=========================================
Our approach to assertions and type checks
=========================================
```

DanielSiepmann (approved):
Thanks for taking the time and solving the issue.

I especially like that you included concrete code examples which makes it easier to grasp and understand as a developer.

Just one small minor RST thing which won't break, that's why I approved already.

Documentation/Assertions.rst
oliverklee: I'm confused … on my screen, your change makes the `===` lines 1 char shorter than the text line, and AFAIK the `===` lines should be the same length. Am I missing something here?

DanielSiepmann: They should. Looks like an issue with GitHub on my end than.
https://raw.githubusercontent.com/FriendsOfTYPO3/tea/3949ed3ea7cfc4ed58ebfb94a2be1c41889e97a4/Documentation/Assertions.rst looks okay.

Screenshot for reference:
image

PR #1357, opened 2024-07-07
[TASK] Move php sniff to build dir

lukaszuznanski:
Part of: https://github.com/FriendsOfTYPO3/tea/issues/1186


oliverklee (changes requested):
We also need to adapt `.gitattributes` and the `prepare-release` Composer script.

composer.json
```diff
@@ -143,7 +143,7 @@
 		],
 		"ci:php:cs-fixer": "php-cs-fixer fix --config .php-cs-fixer.php -v --dry-run --diff",
 		"ci:php:lint": "find .*.php *.php Classes Configuration Tests -name '*.php' -print0 | xargs -r -0 -n 1 -P 4 php -l",
-		"ci:php:sniff": "phpcs Classes Configuration Tests",
+		"ci:php:sniff": "phpcs --standard=Build/phpcs/phpcs.xml Classes Configuration Tests",
```
oliverklee, line 146: We also need to adapt the `fix:php:sniff` Composer script accordingly.

oliverklee (changes requested):

lukaszuznanski:
> We also need to adapt `.gitattributes` and the `prepare-release` Composer script.

We remove whole `Build` dir, so only .gitattributes needs to be adapted.

oliverklee:
> We remove whole `Build` dir, so only `.gitattributes` needs to be adapted.

I was thinking of removing [this line](https://github.com/FriendsOfTYPO3/tea/blob/main/composer.json#L200) that removes a file that will no longer exist after this PR:

```
"rm phpcs.xml",
```

Am I missing something here?

lukaszuznanski:
> > We remove whole `Build` dir, so only `.gitattributes` needs to be adapted.
> 
> I was thinking of removing [this line](https://github.com/FriendsOfTYPO3/tea/blob/main/composer.json#L200) that removes a file that will no longer exist after this PR:
> 
> ```
> "rm phpcs.xml",
> ```
> 
> Am I missing something here?

Yes, that's correct, that needs to be removed.

composer.json
```diff
@@ -147,7 +147,7 @@
 		"ci:php:cs-fixer": "php-cs-fixer fix --config .php-cs-fixer.php -v --dry-run --diff",
 		"ci:php:lint": "parallel-lint .*.php *.php Classes Configuration Tests",
 		"ci:php:mess": "phpmd Classes text Build/phpmd/phpmd.xml",
-		"ci:php:sniff": "phpcs Classes Configuration Tests",
+		"ci:php:sniff": "phpcs --standard=Build/php/phpcs.xml Classes Configuration Tests",
```
oliverklee, line 150:
```suggestion
		"ci:php:sniff": "phpcs --standard=Build/phpcs/phpcs.xml Classes Configuration Tests",
```

oliverklee (changes requested):

oliverklee (approved):

Issue #1113, opened 2024-01-09
Allow installations with Symfony 7


oliverklee:
## Goal
I as a user would like to install Symfony 7 packages
## Acceptance Criteria

- [ ] Symfony 7 packages can be installed

## Additional Information
The TYPO3 Core now allows Symfony 7 on 12LTS and main: https://review.typo3.org/c/Packages/TYPO3.CMS/+/82304

deltanart:
When trying to install the symfony 7 packages I get this error:
```
Your requirements could not be resolved to an installable set of packages.

  Problem 1
    - Root composer.json requires typo3/coding-standards ^0.6.1 -> satisfiable by typo3/coding-standards[v0.6.1].
    - typo3/coding-standards v0.6.1 requires symfony/console ^4.4.30 || ^5.3.7 || ^6.0 -> found symfony/console[v4.4.30, ..., v4.4.49, v5.3.7, ..., v5.4.39, v6.0.0, ..., v6.4.7] but it conflicts with your root composer.json require (^7.0).
```

DanielSiepmann:
I've added the issue to our next team call in order to discuss and solve the situation.

Thanks for investigating.

DanielSiepmann:
There should now be an 0.8.0 release of the CGL package, thanks @oliverklee : https://github.com/TYPO3/coding-standards/issues/125#issuecomment-2145325791

PR #1374, opened 2024-07-29
[TASK] Allow installation of typo3/coding-standards 0.8.x

DanielSiepmann:
This is necessary as older versions block installation of Symfony 7
components.
We also keep 0.6 as 0.8 drops support for older Symfony versions

Resolves: #1113

### DIFF
--- a/composer.json
+++ b/composer.json
@@ -68,7 +68,7 @@
 		"tomasvotruba/cognitive-complexity": "^0.2.3",
 		"tomasvotruba/type-coverage": "^0.3.1",
 		"typo3/cms-fluid-styled-content": "^11.5.4 || ^12.4.2",
-		"typo3/coding-standards": "^0.6 || ^0.8",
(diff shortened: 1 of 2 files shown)


composer.json
```diff
@@ -68,7 +68,7 @@
 		"tomasvotruba/cognitive-complexity": "^0.2.3",
 		"tomasvotruba/type-coverage": "^0.3.1",
 		"typo3/cms-fluid-styled-content": "^11.5.4 || ^12.4.2",
-		"typo3/coding-standards": "^0.6.1",
+		"typo3/coding-standards": "^0.6 || ^0.8",
```
oliverklee, line 71: For the development dependencies, please let's use the exact version:

```suggestion
		"typo3/coding-standards": "^0.6.1 || ^0.8.0",
```

oliverklee (changes requested):

oliverklee (approved):

PR #1376, opened 2024-07-29
[BUGFIX] Dont use parallel execution it is not available in image

Starkmann:


oliverklee:
Closing after discussion: We'll switch to paratest instead.

Issue #1139, opened 2024-01-26
Use `php-parallel-lint/php-parallel-lint` for php linting


simonschaufi:
## Goal
I as a user want to use php-parallel-lint, because it is faster
## Acceptance Criteria

- [ ] php-parallel-lint is implemented

## Additional Information
Suggestion: use `php-parallel-lint/php-parallel-lint` for php linting
https://packagist.org/packages/php-parallel-lint/php-parallel-lint

oliverklee:
Thanks for the suggestion, @simonschaufi!

I've just run the non-parallel PHP lint of my seminars extension with ~120k lines of PHP: PHP-linting all files takes ~2.0 seconds on my (fat) 2021 desktop machine and ~2.5 seconds on my ~2020 Thinkpad when it's running on battery. (Both numbers are fore the second run so the files are already cached.)

Those numbers are "good enough" for me, and I'm not sure having the better performance from parallelizing the linting would warrant the added complexity and the additional tool.

Do you have different numbers, @simonschaufi, or am I missing something here?

simonschaufi:
The amount of php files is so small in the tea extension that you can't really see a big difference.
On a huge extension that I have, it makes a massive difference:

tea way:

```
real    0m8,273s
user    0m22,092s
sys     0m10,683s
```

parallel:
```
real    0m1,066s
user    0m3,122s
sys     0m2,634s
```

oliverklee:
@simonschaufi Out pure curiosity, how big is your extension (in terms of lines of code)?

simonschaufi:
Huge! It's a whole shop application.

https://mattstauffer.com/blog/how-to-count-the-number-of-lines-of-code-in-a-php-project/

```
phploc 7.0.2 by Sebastian Bergmann.

Directories                                         75
Files                                              480

Size
  Lines of Code (LOC)                            91137
  Comment Lines of Code (CLOC)                   20918 (22.95%)
  Non-Comment Lines of Code (NCLOC)              70219 (77.05%)
  Logical Lines of Code (LLOC)                   17058 (18.72%)
    Classes                                      16743 (98.15%)
      Average Class Length                          37
        Minimum Class Length                         0
        Maximum Class Length                       787
      Average Method Length                          3
        Minimum Method Length                        0
        Maximum Method Length                      213
      Average Methods Per Class                      9
        Minimum Methods Per Class                    0
        Maximum Methods Per Class                  241
    Functions                                       73 (0.43%)
[...]

PR #1378, opened 2024-07-29
[FEATURE] Use parallel-lint for command ci:php:lint

Starkmann:
Fixes: #1139

### DIFF
--- a/composer.json
+++ b/composer.json
@@ -143,7 +143,7 @@
 			"@ci:php:stan"
 		],
 		"ci:php:cs-fixer": "php-cs-fixer fix --config .php-cs-fixer.php -v --dry-run --diff",
-		"ci:php:lint": "parallel-lint --exclude .git --exclude .cache --exclude .Build --exclude typo3temp --exclude Documentation --exclude Resources --exclude .gitlab --exclude .github .",
+		"ci:php:lint": "parallel-lint .*.php *.php Classes Configuration Tests",
 		"ci:php:sniff": "phpcs Classes Configuration Tests",
 		"ci:php:stan": "phpstan --no-progress",
 		"ci:static": [
--- a/composer.json
+++ b/composer.json
@@ -143,7 +143,7 @@
 			"@ci:php:stan"
 		],
 		"ci:php:cs-fixer": "php-cs-fixer fix --config .php-cs-fixer.php -v --dry-run --diff",
-		"ci:php:lint": ".Build/bin/parallel-lint --exclude .git --exclude .cache --exclude .Build --exclude typo3temp --exclude Documentation --exclude Resources --exclude .gitlab --exclude .github .",
+		"ci:php:lint": "parallel-lint --exclude .git --exclude .cache --exclude .Build --exclude typo3temp --exclude Documentation --exclude Resources --exclude .gitlab --exclude .github .",
 		"ci:php:sniff": "phpcs Classes Configuration Tests",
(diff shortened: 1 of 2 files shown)


oliverklee (changes requested):
I propose we re-use the existing `ci:php:lint` Composer script. We then also need to add `composer install` and composer caching to the corresponding GitHub CI job.

composer.json
```diff
@@ -142,7 +143,7 @@
 			"@ci:php:stan"
 		],
 		"ci:php:cs-fixer": "php-cs-fixer fix --config .php-cs-fixer.php -v --dry-run --diff",
-		"ci:php:lint": "find .*.php *.php Classes Configuration Tests -name '*.php' -print0 | xargs -r -0 -n 1 -P 4 php -l",
+		"ci:php:lint": ".Build/bin/parallel-lint --exclude .git --exclude .cache --exclude .Build --exclude typo3temp --exclude Documentation --exclude Resources --exclude .gitlab --exclude .github .",
```
oliverklee, line 146:
```suggestion
		"ci:php:lint": "parallel-lint --exclude .git --exclude .cache --exclude .Build --exclude typo3temp --exclude Documentation --exclude Resources --exclude .gitlab --exclude .github .",
```

oliverklee (changes requested):
Please also adapt the PR title to match the new purpose of this PR.

Also, currently the documentation renders with warnings which we'll need to fix.

composer.json
```diff
@@ -142,7 +143,7 @@
 			"@ci:php:stan"
 		],
 		"ci:php:cs-fixer": "php-cs-fixer fix --config .php-cs-fixer.php -v --dry-run --diff",
-		"ci:php:lint": "find .*.php *.php Classes Configuration Tests -name '*.php' -print0 | xargs -r -0 -n 1 -P 4 php -l",
+		"ci:php:lint": "parallel-lint --exclude .git --exclude .cache --exclude .Build --exclude typo3temp --exclude Documentation --exclude Resources --exclude .gitlab --exclude .github .",
```
oliverklee, line 146: I propose we use allowlisting for the file and directories to lint instead of switching to blacklisting (as that would need adjustment every time a tool create a PHP file).

oliverklee (changes requested):

composer.json
Starkmann: parallel-lint does not have whitelist:
https://github.com/php-parallel-lint/PHP-Parallel-Lint?tab=readme-ov-file#command-line-options

Starkmann: And combining multiple calls is not what we want, or?

oliverklee: I understand this line from the documentation that the last part of the call is the directory to check. (I haven't tried this, and I also have not tried whether we could provide multiple directories here.)

Documentation/Development.rst
```diff
@@ -0,0 +1,7 @@
+.. include:: /Includes.rst.txt
+
+.. _dependency-manager:
+
+==================
+Dependency manager
+==================
```
oliverklee, line 7: This file should not be part of this PR.

oliverklee (changes requested):

composer.json
oliverklee, line 146:
```suggestion
		"ci:php:lint": "parallel-lint .*.php *.php Classes Configuration Tests",
```

oliverklee (approved):

Issue #1379, opened 2024-07-29
Use paratest for parallel test execution


Starkmann:
https://github.com/paratestphp/paratest

Starkmann:
Do we still need this (build/docker_install.sh):
`apk add parallel`

PR #1381, opened 2024-07-29
[FEATURE] Use paratest for composer script

Starkmann:
This needs to be done for runTests.sh as well

Resolves: #1379


oliverklee (changes requested):
This change also needs a rebase.

Please also update the commit message and PR description from `Resolves: #xxx` to `Part of #xxx` to avoid auto-resolving the ticket as this PR only solves part of it.

Starkmann:
With parallel execution we have a race condition between functional test since they all are running on the same database:

Duplicate entry '1' for key 'tx_tea_domain_model_tea.PRIMARY'

oliverklee:
@Starkmann This PR still has an open review comment.

composer.json
```diff
@@ -166,7 +167,7 @@
 		"ci:tests:create-directories": "mkdir -p .Build/public/typo3temp/var/tests",
 		"ci:tests:functional": [
 			"@ci:tests:create-directories",
-			"find 'Tests/Functional' -wholename '*Test.php' | parallel --gnu 'echo; echo \\\"Running functional test suite {}\\\"; .Build/bin/phpunit -c Build/phpunit/FunctionalTests.xml {}';"
+			"paratest -c Build/phpunit/FunctionalTests.xml Tests/Functional"
 		],
 		"ci:tests:unit": "phpunit -c Build/phpunit/UnitTests.xml Tests/Unit",
```
oliverklee, line 172: This probably should be `paratest` as well.

oliverklee (changes requested):

oliverklee (approved):

PR #1383, opened 2024-07-29
[TASK] Remove no longer needed path to binaries

Starkmann:


composer.json
```diff
@@ -159,9 +159,9 @@
 		"ci:tests:create-directories": "mkdir -p .Build/public/typo3temp/var/tests",
 		"ci:tests:functional": [
 			"@ci:tests:create-directories",
-			"find 'Tests/Functional' -wholename '*Test.php' | parallel --gnu 'echo; echo \\\"Running functional test suite {}\\\"; .Build/bin/phpunit -c Build/phpunit/FunctionalTests.xml {}';"
+			"find 'Tests/Functional' -wholename '*Test.php' | parallel --gnu 'echo; echo \\\"Running functional test suite {}\\\"; phpunit -c Build/phpunit/FunctionalTests.xml {}';"
```
oliverklee, line 162: For the cases where we are piping something to a Composer script (in this case: functional tests), we'll need to keep using the full path AFAIK.

oliverklee (changes requested):

oliverklee (changes requested):
Also, in the commit message and PR title, this one here should be a `[TASK]`, not a `[FEATURE]`.

composer.json
Starkmann: Might be can not really test this because of parallel error

oliverklee (commented):
Looks good, but needs a rebase to resolve the merge conflicts.

Issue #447, opened 2022-05-15
Add `phpcompatibility/php-compatibility` to the toolchain


oliverklee:
This is a set of rules for PHP_CodeSniffer that checks for compatibility with different versions of PHP.

It is suggested to use the development branch, not one of the released versions.

oliverklee:
We have failed to get this to work and have decided that this tool isn't worth the hassle at the moment. Closing.

PR #1387, opened 2024-07-29
[FEATURE] Add PHP compatibility sniffer

Starkmann:
Resolves: #447

### DIFF
--- a/composer.json
+++ b/composer.json
@@ -52,6 +52,7 @@
 		"helmich/typo3-typoscript-lint": "^3.1.1",
 		"icanhazstring/composer-unused": "^0.8.11",
 		"php-coveralls/php-coveralls": "^2.7.0",
+		"phpcompatibility/php-compatibility": "^9.3",
 		"phpstan/extension-installer": "^1.4.1",
 		"phpstan/phpstan": "^1.11.8",
 		"phpstan/phpstan-phpunit": "^1.4.0",
@@ -116,6 +117,8 @@
 		}
 	},
 	"scripts": {
+		"post-install-cmd": "phpcs --config-set installed_paths vendor/phpcompatibility/php-compatibility",
+		"post-update-cmd" : "phpcs --config-set installed_paths vendor/phpcompatibility/php-compatibility",
 		"ci:composer:normalize": "@composer normalize --no-check-lock --dry-run",
 		"ci:composer:psr-verify": "@composer dumpautoload --optimize --strict-psr --no-plugins",
 		"ci:composer:unused": "composer-unused",
--- a/composer.json
+++ b/composer.json
@@ -52,7 +52,7 @@
 		"helmich/typo3-typoscript-lint": "^3.1.1",
 		"icanhazstring/composer-unused": "^0.8.11",
 		"php-coveralls/php-coveralls": "^2.7.0",
-		"phpcompatibility/php-compatibility": "^9.3",
+		"phpcompatibility/php-compatibility": "^9.3.5",
 		"phpstan/extension-installer": "^1.4.1",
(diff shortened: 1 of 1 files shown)


composer.json
```diff
@@ -143,7 +144,10 @@
 		],
 		"ci:php:cs-fixer": "php-cs-fixer fix --config .php-cs-fixer.php -v --dry-run --diff",
 		"ci:php:lint": "find .*.php *.php Classes Configuration Tests -name '*.php' -print0 | xargs -r -0 -n 1 -P 4 php -l",
-		"ci:php:sniff": "phpcs Classes Configuration Tests",
+		"ci:php:sniff": [
+			"phpcs --config-set installed_paths vendor/phpcompatibility/php-compatibility",
```
oliverklee, line 148: This needs to be `.Build/vendor/…`.

oliverklee: According to the docs, it's recommended to use a Composer plugin to register the additional sniffs. https://github.com/PHPCompatibility/PHPCompatibility?tab=readme-ov-file#installation-in-a-composer-project-method-1

composer.json
```diff
@@ -51,6 +51,7 @@
 		"friendsofphp/php-cs-fixer": "^3.60.0",
 		"helmich/typo3-typoscript-lint": "^3.1.1",
 		"php-coveralls/php-coveralls": "^2.7.0",
+		"phpcompatibility/php-compatibility": "^9.3",
```
oliverklee, line 54: For development dependencies, please always use the exact patch-level version (as we don't risk breaking anything with this):

```suggestion
		"phpcompatibility/php-compatibility": "^9.3.5",
```

oliverklee (changes requested):

composer.json
```diff
@@ -96,7 +97,8 @@
 			"phpstan/extension-installer": true,
 			"sbuerk/typo3-cmscomposerinstallers-testingframework-bridge": true,
 			"typo3/class-alias-loader": true,
-			"typo3/cms-composer-installers": true
+			"typo3/cms-composer-installers": true,
+			"dealerdirect/phpcodesniffer-composer-installer": true
```
oliverklee, line 101: Please let's also keep these entries sorted in alphabetical order.

oliverklee (changes requested):

oliverklee (changes requested):
Also, please run `./Build(/Scripts/runTests.sh composer normalize` to auto-order some things in the `composer.json`.

composer.json
```diff
@@ -143,7 +146,10 @@
 		],
 		"ci:php:cs-fixer": "php-cs-fixer fix --config .php-cs-fixer.php -v --dry-run --diff",
 		"ci:php:lint": "find .*.php *.php Classes Configuration Tests -name '*.php' -print0 | xargs -r -0 -n 1 -P 4 php -l",
-		"ci:php:sniff": "phpcs Classes Configuration Tests",
+		"ci:php:sniff": [
+			"phpcs --config-set installed_paths .Build/vendor/phpcompatibility/php-compatibility",
```
oliverklee, line 150: Is this line still needed now that we're using the plugin?

oliverklee (changes requested):
We also should configure the PHP version range that this check should check for.

From the documentation:

> You can also specify a range of PHP versions that your code needs to support. In this situation, compatibility issues that affect any of the PHP versions in that range will be reported: `--runtime-set testVersion 5.3-5.5`

oliverklee (changes requested):
We also need to resolve the merge conflicts.

oliverklee:
We have failed to get this to work and have decided that this tool isn't worth the hassle at the moment. Closing.

PR #1388, opened 2024-07-29
[DOCS] Extend PHIVE section in docs

### LINKED ISSUE #429: Document why we're using phpcov via PHIVE
## Goal
I as a developer want to document why we are using phpcov via PHIVE
## Acceptance Criteria

- [ ] the documentation is written
## Additional Information
https://github.com/sebastianbergmann/phpcov

### LINKED ISSUE #543: Add a small section to the docs with the PHIVE usage essentials
## Goal
I as a user need the information, why the PHIVE usage is essential
## Acceptance Criteria

- [ ] the information ist documented in the docs

kanow:
Fixes #543
Fixes #429

### DIFF
--- a/Documentation/DependencyManager.rst
+++ b/Documentation/DependencyManager.rst
@@ -16,3 +16,18 @@ install or upgrade some tool as the tool's dependencies conflict with the
 dependencies on another library). It also allows running versions of tools
 that require a PHP version that is higher than the lowest allowed PHP version
 for this project.
+
+.. _using-phive-to-install-phpcov:
+
+Using PHIVE to install `phpunit/phpcov`
+=======================================
+
+To support php version 7.4 and 8.2 in the `tea` extension, we are using PHIVE
+to install `phpunit/phpcov`.
+We need `phpunit/phpcov` in version 10 to support php 8.2.
+Our minimum php version 7.4 would prevent the installation with composer.
+
+.. note::
+
+   To find more information about install and usage,
+   please check out the documentation of `PHIVE <https://phar.io>`__.


Documentation/DependencyManager.rst
```diff
@@ -16,3 +16,18 @@ install or upgrade some tool as the tool's dependencies conflict with the
...
+
+.. _using-phive-to-install-phpcov:
+
+Using PHIVE to install `phpunit/phpcov`
+=======================================
+
+To support php version 7.4 and 8.2 in the `tea` extension, we are using PHIVE
+to install `phpunit/phpcov`.
+We need `phpunit/phpcov` in version 10 to support php 8.2.
+Our minimum php version 7.4 would prevent the installation with composer.
```
oliverklee, line 28: I'd suggest to explain this in more detail to explain the problem, e.g., with this wording:

`To keep things simple, we try do model as many dependencies as possible as Composer dependencies in the `composer.json`. However, sometime a dependency is not available for all PHP versions that we support, or it might conflict with another dependency. One of this cases is PHPCOV, a library we use to collect and transmit code coverage information: The version that works with PHP 8.2 and 8.3 does not work with PHP 7.3. So we have switch this particular dependency from being installed via Composer to being installed via PHIVE.

oliverklee (changes requested):

Documentation/DependencyManager.rst
oliverklee: We could also add this explanation:

As we usually only collect code coverage data on CI and not locally, it's usually not necessary to install PHPDOV locally.

oliverklee:
@kanow Let's get this PR finished and merged! :rocket:

oliverklee (approved):

PR #1396, opened 2024-07-30
[FEATURE] Allow running npm based quality checks via `runTests.sh`

cumuru:
Additionally, a generic npm command has been added. 
See `runTests.sh --help` to explore the new commands `npm`, `lintStyle` and `lintJs`

Fixes #972

PR #1406, opened 2024-07-30
[TASK] Rename the CSS lint npm command

oliverklee:
The name of the command should reflect the type of file that gets linted.

This is a pre-patch to #1396.


cumuru (approved):
Thanks for separating!

Issue #1312, opened 2024-05-14
`runTests.sh` leaves Docker networks around


oliverklee:
I as a user want to shutdown the networks after the `runTests.sh` command is finished.

## Acceptance criteria

- [ ] when the `runTests.sh` command is finished all related networks are shutdown

## Additional Information
This is after restarting my machine.

```
klee@gonzales$ docker network ls
NETWORK ID     NAME                               DRIVER    SCOPE
6920792e352c   bridge                             bridge    local
4198d7d220b4   friendsoftypo3-tea-660             bridge    local
084e52b5790f   friendsoftypo3-tea-2942            bridge    local
fc98aedbb1e8   friendsoftypo3-tea-5782            bridge    local
b0147ccd1aff   friendsoftypo3-tea-13929           bridge    local
15566258b4f5   friendsoftypo3-tea-19349           bridge    local
0d9d6743a587   friendsoftypo3-tea-19585           bridge    local
13e44a16e2dc   friendsoftypo3-tea-22532           bridge    local
e08253a9b4d2   friendsoftypo3-tea-23251           bridge    local
c3a4aa8fb79d   friendsoftypo3-tea-24302           bridge    local
7c265280ddc6   friendsoftypo3-tea-29587           bridge    local
1e2b7e72979f   host                               host      local
f2bc4af41dd4   none                               null      local
```

This causes problems for DDEV (from another project) as DDEV then cannot find a free network range and refuses to start:

>  Error response from daemon: could not find an available, non-overlapping IPv4 address pool among the defaults to assign to the network' 

I would expect the networks to be gone after the `runTests.sh` command is finished.

oliverklee:
Possible solution: https://github.com/TYPO3/testing-framework/pull/584

Starkmann:
I can reproduce this behavior for commands that are not existing:
`./Build/Scripts/runTests.sh -s functionalxxxx
`

PR #1411, opened 2024-07-31
[BUGFIX] Clean up network after invalid command

Starkmann:
Related: #1312

### DIFF
--- a/Build/Scripts/runTests.sh
+++ b/Build/Scripts/runTests.sh
@@ -658,7 +658,11 @@ case ${TEST_SUITE} in
         echo "Invalid -s option argument ${TEST_SUITE}" >&2
         echo >&2
         echo "${HELP}" >&2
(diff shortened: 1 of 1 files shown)


Build/Scripts/runTests.sh
```diff
@@ -649,6 +649,7 @@ case ${TEST_SUITE} in
         echo "Invalid -s option argument ${TEST_SUITE}" >&2
         echo >&2
         echo "${HELP}" >&2
+        ${CONTAINER_BIN} network remove ${NETWORK} >/dev/null
```
oliverklee, line 652: I think we need to have the same Docker/Podman switch here that the Testing framework has:

https://github.com/TYPO3/testing-framework/pull/584/files

oliverklee (changes requested):

oliverklee (approved):
Looks good to me!

Issue #1412, opened 2024-07-31
Ditch `runTests.sh composerInstall`


oliverklee:
## Background

Currently, we have two Composer-related commands in `runTests.sh`, which both are copied from the Core `runTests.sh`:
- the older `composerInstall` command, which can relay additional options only after providing `---`
- the newer `composer` wrapper, which can pass additional options without the requirement of passing `--`

The Core has a `composer.lock` (#1161), and hence running `composer install` makes sense there. We don't have a `composer.lock`, and hence we should be using `composer update` instead.

## Decision to take

Drop the `composerInstall` command from `runTests.sh`?

## Additional benefits

Then we also could avoid having to fix options for `composerInstall`.

DanielSiepmann:
We decided to remove the alias.
That way we do not have a wrapper and people can use `-s composer` as they expect.
Also there is no `composer.lock` so `install` was wrong anyway and should be `update`.

oliverklee:
Fixed by #1416.

PR #1416, opened 2024-07-31
[TASK] Remove undocumented composerInstall

DanielSiepmann:
The project doesn't have a `composer.lock`. That's why one should use `composer update` instead of `composer lock`.
This can be done by using the existing `-s composer` and appending `update.
Furthermore, the command itself was not documented or used anyway.

Relates: #1412


oliverklee:
This change needs a rebase to resolve the merge conflicts.

oliverklee (approved):
Looks good.

PR #1417, opened 2024-07-31
[TASK] Rename composerInstall* scripts

DanielSiepmann:
`composer install` does not make sense as we do not have a composer.lock. One should use `composer update` instead. This is reflected by updating the script names.

Relates: #1412


oliverklee (approved):
Looks good to me.

PR #1419, opened 2024-07-31
[TASK] Add functional tests for `showAction`

### LINKED ISSUE #1228: Add a functional test for the Tea single view

Starkmann:
Fixes #1228

### DIFF
--- a/Tests/Functional/Controller/TeaControllerTest.php
+++ b/Tests/Functional/Controller/TeaControllerTest.php
@@ -68,7 +68,7 @@ final class TeaControllerTest extends FunctionalTestCase
      */
     public function showActionRendersTheGivenTeas(): void
     {
-        $request = (new InternalRequest())->withPageId(1)->withQueryParameters(["tx_tea_teashow[tea]" => 1]);
+        $request = (new InternalRequest())->withPageId(1)->withQueryParameters(['tx_tea_teashow[tea]' => 1]);
 
         $html = (string)$this->executeFrontendSubRequest($request)->getBody();
 
--- a/Tests/Functional/Controller/TeaControllerTest.php
+++ b/Tests/Functional/Controller/TeaControllerTest.php
@@ -4,6 +4,7 @@ declare(strict_types=1);
 
 namespace TTN\Tea\Tests\Functional\Controller;
 
+use TYPO3\CMS\Core\Utility\ArrayUtility;
 use TYPO3\TestingFramework\Core\Functional\Framework\Frontend\InternalRequest;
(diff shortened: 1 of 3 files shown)


Tests/Functional/Controller/TeaControllerTest.php
```diff
@@ -22,6 +23,14 @@ final class TeaControllerTest extends FunctionalTestCase
 
     protected function setUp(): void
     {
+        ArrayUtility::mergeRecursiveWithOverrule($this->configurationToUseInTestInstance, [
+            'FE' => [
+                'cacheHash' => [
+                    'enforceValidation' => false,
```
oliverklee, line 29: Could we also use a URL with a cHash in the test instead?

oliverklee: Or maybe configure human-readable URLs in the site configuration?

Tests/Functional/Controller/TeaControllerTest.php
```diff
@@ -55,4 +63,21 @@ public function indexActionRendersAllAvailableTeas(): void
...
+
+    /**
+     * @test
+     */
+    public function showActionRendersTheGivenTeas(): void
+    {
+        $request = new InternalRequest();
+        $request = $request->withPageId(3);
+        $request = $request->withQueryParameters([
+            "tx_tea_teashow[tea]" => 1
```
oliverklee, line 75: These should be single quotes.

oliverklee (changes requested):
The test generally looks good. There are some style warnings, though, and the branch needs a rebase.

Also, I'd prefer a cleaner solution over disabling cHash validation.

Tests/Functional/Controller/TeaControllerTest.php
Starkmann: We do not have site configuration right now in functional test and we would also nee route enhancer for this.

Tests/Functional/Controller/TeaControllerTest.php
Starkmann: Done

Tests/Functional/Controller/TeaControllerTest.php
oliverklee: So would it be possible to create/use a valid cHash here?

oliverklee: Done (without cHash, but with a simpler setup).

oliverklee (approved):

oliverklee (changes requested):
I'll need to have a look at the lowest test failures.

oliverklee (approved):

PR #1422, opened 2024-08-01
[TASK] Polish the `runTests.sh` help

oliverklee:
- sort some lists
- use the proper capitalization for PHPStan


cumuru (changes requested):
There are four commands that are not documented:
* composerInstall (should have gone with #1417)
* composerNormalize 
* npm
* update
I suggest to either add to the help text or drop them.
Thanks!

oliverklee:
> There are four commands that are not documented:

I'd like to add those in a separate PR.

cumuru (approved):

oliverklee:
> There are four commands that are not documented:

Done: #1423

PR #1430, opened 2024-08-23
[TASK] Switch to the Coveralls GitHub action

oliverklee:
Unfortunately, with the GitHub Action, we need to go back to using the path `build/logs/clover.xml` for the combined coverage file.

Closes #1431

### DIFF
--- a/.github/workflows/codecoverage.yml
+++ b/.github/workflows/codecoverage.yml
@@ -75,12 +75,15 @@ jobs:
         run: "ls -lahR .Build/coverage/"
       - name: "Merge coverage results"
         run: composer ci:coverage:merge
-      - name: "Show generated coverage files"
-        run: "ls -lahR .Build/coverage/"
+      - name: "Show combined coverage files"
+        run: "ls -lahR build/logs/"
       - name: "Upload coverage results to Coveralls"
+        uses: coverallsapp/github-action@v2
         env:
-          COVERALLS_REPO_TOKEN: ${{ secrets.GITHUB_TOKEN }}
-        run: .Build/bin/php-coveralls --coverage_clover=.Build/coverage/combined.xml --json_path=.Build/coverage/coveralls-upload.json -v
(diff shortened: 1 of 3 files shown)


.github/workflows/codecoverage.yml
```diff
@@ -75,12 +75,15 @@ jobs:
...
+        run: "ls -lahR build/logs/"
       - name: "Upload coverage results to Coveralls"
+        uses: coverallsapp/github-action@v2
         env:
-          COVERALLS_REPO_TOKEN: ${{ secrets.GITHUB_TOKEN }}
-        run: .Build/bin/php-coveralls --coverage_clover=.Build/coverage/combined.xml --json_path=.Build/coverage/coveralls-upload.json -v
+          github-token: ${{ secrets.GITHUB_TOKEN }}
+          # Note: This is the only path that the Coveralls GitHub Action supports.
+          # So we cannot use something like .Build/coverage/clover.xml here.
+          file: build/logs/clover.xml
```
DanielSiepmann, line 86: Can you share the source? https://github.com/coverallsapp/github-action mentiones:

> 
> 
> file 	optional 	Default: all coverage files that could be found. Local path to the coverage report file produced by your test suite. An error will be thrown if no file was found. This is the file that will be sent to the Coveralls API. Leave empty if you want to combine many files reports.
> files 	optional 	Default: all coverage files that could be found. Space-separated list of coverage report files produced by your test suite. Example: files: coverage/test1.lcov coverage/test2.lcov

Which would allow to define the file.

oliverklee: Yes, according to the docs, this should work. In practice, it doesn't work if I change the file name or the directory. Also, it doesn't work if I use `files` instead of `file`. You can see this in this 3 demo PRs:
- #1433
- #1434
- #1435

DanielSiepmann (approved):

PR #1468, opened 2024-09-21
[TASK] Bump the Symfony development dependencies

oliverklee:
For development versions, we don't need to bother supporting lower versions of dependencies.


oliverklee:
@DanielSiepmann @lukaszuznanski gentle ping for a review :-)

DanielSiepmann (approved):

DanielSiepmann (approved):
@oliverklee approved if my question would be answered with no: Doesn't that prevent PHPStan, PHPUnit, etc. runs within CI against the lowest possible runtime dependencies, as require-dev constraints would raise the minimum version? And would you see this as an issue?

oliverklee:
@DanielSiepmann Yes, this change has affects on which versions of those dependencies get installed when we install the lowest dependencies for unit and functional tests.

Those are development dependencies: We're not using them directly in production code (although the TYPO3 Core code might possibly call them), but (in this case) only in CI jobs like XLIFF and YAML linting. So I don't think raising the dependencies would be an issue in such a way that it would keep us from catching real problems.

PR #1484, opened 2024-10-03
[TASK] Move a repository test from unit to functional

oliverklee:
Unit tests for repositories are quite a hassle as we then need to set up all dependencies in `setUp()` ourselves.

So convert the test to a functional tests where we can rely on the container to create and initialize the subject for us.


Tests/Functional/Domain/Repository/TeaRepositoryTest.php
```diff
@@ -31,6 +32,14 @@ protected function setUp(): void
         $this->subject = $this->get(TeaRepository::class);
     }
 
+    /**
+     * @test
+     */
+    public function isRepository(): void
```
DanielSiepmann, line 38: I don't see a reason to add this method. Can you explain?

DanielSiepmann (approved):

Tests/Functional/Domain/Repository/TeaRepositoryTest.php
oliverklee: The rationale for the existence of this test (not for moving it) is to test that our repository extends the base class and hence can be trusted to have all the behavior of the base class (which hopefully is test-covered in Extbase). Otherwise, we'd need to have tests for the base class methods to cover the full code of the class.

PR #1536, opened 2025-01-03
[FEATURE] Add support for PHP 8.4

oliverklee:


.github/workflows/ci.yml
```diff
@@ -175,6 +176,12 @@ jobs:
           - typo3-version: "^11.5"
             php-version: "8.3"
             composer-dependencies: highest
+#          - typo3-version: "^11.5"
+#            php-version: "8.4"
+#            composer-dependencies: lowest
+          - typo3-version: "^11.5"
+            php-version: "8.4"
+            composer-dependencies: highest
```
DanielSiepmann, line 184: We just removed support for v11.
Can you rebase and resolve the conflicts?

.github/workflows/ci.yml
```diff
@@ -252,6 +259,12 @@ jobs:
           - typo3-version: "^12.4"
             php-version: "8.3"
             composer-dependencies: highest
+#          - typo3-version: "^12.4"
+#            php-version: "8.4"
+#            composer-dependencies: lowest
```
DanielSiepmann, line 264: Why do we keep this commented out?

.github/workflows/ci.yml
```diff
@@ -175,6 +176,12 @@ jobs:
           - typo3-version: "^12.4"
             php-version: "8.3"
             composer-dependencies: highest
+          - typo3-version: "^12.4"
```
DanielSiepmann, line 179: Don't we need those new jobs also within GitLab?

.github/workflows/ci.yml
oliverklee: Because it fails: https://github.com/TYPO3BestPractices/tea/actions/runs/13112021223

.github/workflows/ci.yml
oliverklee: Yes, on it.

.github/workflows/ci.yml
oliverklee: This is due to an incompatible Symfony DI version, and requiring a higher version leads to problems in other components, which then also would need to be required in higher versions etc. etc., which opens a whole can of worms.

DanielSiepmann: Ah okay … yeah fine for me for now, as highest is working already.

DanielSiepmann (approved):

Issue #1566, opened 2025-02-03
Drop support for PHP < 8.1


oliverklee:
## Goal

In preparation for dropping support for TYPO3 11LTS, we'd like to drop support for PHP < 8.1.

## Acceptance Criteria

- [x] The minimum PHP requirement in `composer.json` and `ext_emconf.php` is PHP 8.1.
- [x] All CI jobs (GitHub Actions) for PHP < 8.1 are gone.
- [x] All CI jobs (GitLab Ci) for PHP < 8.1 are gone.
- [x] The Rector and PHPStan configuration are still unchanged at this moment (so the language level is still PHP 7.4).

Issue #1568, opened 2025-02-03
Raise the language level to PHP 8.1


oliverklee:
## Goal

Let's use the power and features of newer PHP versions!

## Acceptance Criteria

- [x] The Rector and PHPStan configuration are set to PHP >= 8.1 (if required/possible).
- [x] We use the latest PHP 8.1 features in our code.

PR #1576, opened 2025-02-03
[!!!][TASK] Drop support for PHP < 8.1

DanielSiepmann:
As this is not supported by current TYPO3 LTS versions.

Resolves: #1566


rector.php
```diff
@@ -26,7 +26,7 @@
         __DIR__ . '/ext_emconf.php',
         __DIR__ . '/ext_localconf.php',
     ])
-    ->withPhpVersion(PhpVersion::PHP_74)
+    ->withPhpVersion(PhpVersion::PHP_81)
```
oliverklee, line 29: Let's keep this at 7.4 for the time being and then change it as part of #1568.

oliverklee (changes requested):

PR #1577, opened 2025-02-03
[TASK] Raise PHP Language Level to 8.1

DanielSiepmann:
Resolves: #1568

### DIFF
--- a/Classes/Controller/FrontEndEditorController.php
+++ b/Classes/Controller/FrontEndEditorController.php
@@ -17,15 +17,10 @@ use TYPO3\CMS\Extbase\Mvc\Controller\ActionController;
  */
 class FrontEndEditorController extends ActionController
 {
-    private Context $context;
-
-    private TeaRepository $teaRepository;
-
-    public function __construct(Context $context, TeaRepository $teaRepository)
-    {
-        $this->context = $context;
-        $this->teaRepository = $teaRepository;
-    }
+    public function __construct(
+        private readonly Context $context,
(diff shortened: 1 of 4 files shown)


DanielSiepmann:
The php Code Sniffer and php-cs-fixer have different rules.
I'd opt for removing the PHP Code Sniffer in favour of php-cs-fixer and only use one tool for CGL.

Classes/Controller/FrontEndEditorController.php
```diff
@@ -17,15 +17,10 @@
...
-
-    public function __construct(Context $context, TeaRepository $teaRepository)
-    {
-        $this->context = $context;
-        $this->teaRepository = $teaRepository;
-    }
+    public function __construct(
+        private readonly Context $context,
+        private readonly TeaRepository $teaRepository,
+    ) {}
```
DanielSiepmann, line 23: The php Code Sniffer and php-cs-fixer have different rules.
I'd opt for removing the PHP Code Sniffer in favour of php-cs-fixer and only use one tool for CGL.

oliverklee: Let's do that in a separate PR though, please.

oliverklee:
I have created #1580 for removing PHP_CodeSniffer.

oliverklee (changes requested):
Rector now wants to add some more changes which we should adopt.

oliverklee (approved):

Issue #1567, opened 2025-02-03
Drop support for TYPO3 11LTS


oliverklee:
## Goal

We want to drop support for TYPO3 11LTS as it has reached the end of its public support, and to allow adding support for TYPO3 13LTS later.

## Acceptance Criteria

- [x] The minimum supported TYPO3 version in `composer.json` and `ext_emconf.php` is TYPO3 12.4.26 (the current latest maintenance release).
- [x] All CI jobs for TYPO3 11LTS for GitHub actions are gone.
- [x] All CI jobs for TYPO3 11LTS for GitLab CI are gone.
- [x] All TYPO3-11LTS-specific code is gone. Version-specific code switches are also gone.
- [x] Rector and PHPStan are configured for TYPO3 12LTS.
- [ ] The 11LTS badge is gone from the README.

PR #1578, opened 2025-02-03
[!!!][TASK] Remove TYPO3 v11 Support

DanielSiepmann:
As official v11 support has stopped.

Resolves: #1567


oliverklee (changes requested):
We also should remove the 11LTS badge from the README.

oliverklee (approved):

PR #1592, opened 2025-02-04
[BUGFIX] Improve type safety for lazy-loaded associations

oliverklee:
`_loadRealInstance` will return `null` if the target of the association does not exist (anymore). We need to handle that case in a clean way.

Also use a dedicated folder for the fixture to make the fixture easier to find, and to discourage reuse of fixtures.

### DIFF
--- a/Classes/Domain/Model/Tea.php
+++ b/Classes/Domain/Model/Tea.php
@@ -52,9 +52,8 @@ class Tea extends AbstractEntity
     public function getImage(): ?FileReference
     {
         if ($this->image instanceof LazyLoadingProxy) {
-            /** @var FileReference $image */
             $image = $this->image->_loadRealInstance();
-            $this->image = $image;
+            $this->image = ($image instanceof FileReference) ? $image : null;
         }
 
         return $this->image;
--- a/Tests/Functional/Domain/Repository/Fixtures/propertyMapping/TeaWithDeletedImage.csv
+++ b/Tests/Functional/Domain/Repository/Fixtures/propertyMapping/TeaWithDeletedImage.csv
@@ -1,0 +1,3 @@
+"tx_tea_domain_model_tea"
+,"uid","title","image",
+,1,"Gunpowder",1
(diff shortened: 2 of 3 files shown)


DanielSiepmann:
> `_loadRealInstance` will return `null` if the target of the association does not exist (anymore). We need to handle that case.

This confuses me. We will handle that situation by using `null`, but you wrote that it would already return `null`. So why do we need condition which would do the same?

Classes/Domain/Model/Tea.php
```diff
@@ -52,9 +52,8 @@ public function setDescription(string $description): void
     public function getImage(): ?FileReference
     {
         if ($this->image instanceof LazyLoadingProxy) {
-            /** @var FileReference $image */
             $image = $this->image->_loadRealInstance();
-            $this->image = $image;
+            $this->image = ($image instanceof FileReference) ? $image : null;
```
DanielSiepmann, line 56: > `_loadRealInstance` will return `null` if the target of the association does not exist (anymore). We need to handle that case.

This confuses me. We will handle that situation by using `null`, but you wrote that it would already return `null`. So why do we need condition which would do the same?

oliverklee: To begin, the original type annotation was incorrect:

```
/** @var FileReference $image */
```

In addition, generally type safety via PHP code instead of annotations is to be preferred (as it's technically enforced). As PHPStan does not know (and currently cannot know) the types that `_loadRealInstance` can return, I chose to make the possibilities including the fallback explicit.

DanielSiepmann: Thanks for clarification. So the actual change is for PHPStan, to understand the actual types. It is not for actual code execution. Now the change makes sense to me. But that doesn't match the title and description, does it?

oliverklee: I have updated the title and description accordingly.

DanielSiepmann (approved):

PR #1631, opened 2025-04-05
[BUGFIX] Delete current configuration files with `prepare-release`

oliverklee:


oliverklee:
The failure that this PR fixes can be seen in #1630.

DanielSiepmann (approved):

Issue #1565, opened 2025-02-03
Feature: Add functional Tests for TeaController


DanielSiepmann:
## Goal

Provide concrete examples on how to test frontend controllers rendering data from the database.

## Acceptance Criteria
- [x] Test covering an index with no teas. Solved via #1642
- [x] Test covering an index with two teas. Solved via: #21
- [x] Test covering the show action with an available tea. Solved via: #1228, #1419
- [x] Test covering the show action with an invalid tea. Solved via: #1643

## Additional Information
Anything else that would be helpful to know to implement this feature.

PR #1643, opened 2025-04-07
[FEATURE] Add proper 404 handling for tea detail action

d-s-codappix:
Properly trigger TYPO3 404 handling in case a none available tea was requested.
That might happen when old links to no longer existing teas are called. Or if people play around with the URL.

We also cover this new feature with a functional test.

Related: #1565

### DIFF
--- a/Classes/Controller/TeaController.php
+++ b/Classes/Controller/TeaController.php
@@ -27,6 +27,9 @@ class TeaController extends ActionController
         return $this->htmlResponse();
     }
 
+    /**
+     * @throws PropagateResponseException
+     */
     public function showAction(?Tea $tea = null): ResponseInterface
     {
         if ($tea === null) {
@@ -39,6 +42,8 @@ class TeaController extends ActionController
 
     /**
      * Will throw exception to trigger 404.
+     *
+     * @throws PropagateResponseException
      *
      * @return never
      */
--- a/Classes/Controller/TeaController.php
+++ b/Classes/Controller/TeaController.php
@@ -7,7 +7,9 @@ namespace TTN\Tea\Controller;
 use Psr\Http\Message\ResponseInterface;
 use TTN\Tea\Domain\Model\Tea;
 use TTN\Tea\Domain\Repository\TeaRepository;
+use TYPO3\CMS\Core\Http\PropagateResponseException;
 use TYPO3\CMS\Extbase\Mvc\Controller\ActionController;
+use TYPO3\CMS\Frontend\Controller\ErrorController;
 
 /**
  * Controller for the main "Tea" FE plugin.
@@ -16,6 +18,7 @@ class TeaController extends ActionController
 {
     public function __construct(
         private readonly TeaRepository $teaRepository,
+        private readonly ErrorController $errorController,
     ) {}
 
     public function indexAction(): ResponseInterface
@@ -24,9 +27,26 @@ class TeaController extends ActionController
         return $this->htmlResponse();
     }
 
-    public function showAction(Tea $tea): ResponseInterface
+    public function showAction(?Tea $tea = null): ResponseInterface
     {
+        if ($tea === null) {
+            $this->trigger404('No tea given.');
+        }
+
         $this->view->assign('tea', $tea);
         return $this->htmlResponse();
+    }
+
+    /**
+     * Will throw exception to trigger 404.
+     *
+     * @return never
+     */
+    protected function trigger404(string $message): void
+    {
+        throw new PropagateResponseException(
+            $this->errorController->pageNotFoundAction($this->request, $message),
+            1744021673
+        );
     }
 }
--- a/Classes/Controller/TeaController.php
+++ b/Classes/Controller/TeaController.php
@@ -41,7 +41,7 @@ class TeaController extends ActionController
     }
 
     /**
-     * Will throw exception to trigger 404.
+     * @throws PropagateResponseException
      *
      * @throws PropagateResponseException
      *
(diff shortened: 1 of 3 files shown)


Tests/Unit/Controller/TeaControllerTest.php
```diff
@@ -1,107 +0,0 @@
-<?php
```
oliverklee, line 1: This change belongs in #1644, not in this PR.

Classes/Controller/TeaController.php
```diff
@@ -24,9 +27,26 @@ public function indexAction(): ResponseInterface
...
         $this->view->assign('tea', $tea);
         return $this->htmlResponse();
     }
+
+    /**
+     * Will throw exception to trigger 404.
+     *
+     * @return never
+     */
+    protected function trigger404(string $message): void
```
oliverklee, line 45: We should also add an `@throws` annotation.

Tests/Functional/Controller/TeaControllerTest.php
```diff
@@ -67,11 +68,25 @@ public function indexActionRendersAllAvailableTeas(): void
...
         self::assertStringContainsString('Godesberger Burgtee', $html);
         self::assertStringNotContainsString('Oolong', $html);
     }
+
+    /**
+     * @test
+     */
+    public function showActionTriggers404ForMissingTea(): void
+    {
+        $request = (new InternalRequest())->withPageId(3)->withQueryParameters(['tx_tea_teashow[tea]' => 1]);
```
oliverklee, line 86: For this test, we don't need to query parameters AFAICS.

Tests/Functional/Controller/TeaControllerTest.php
```diff
@@ -67,11 +68,25 @@ public function indexActionRendersAllAvailableTeas(): void
...
         $html = (string)$this->executeFrontendSubRequest($request)->getBody();
 
         self::assertStringContainsString('Godesberger Burgtee', $html);
         self::assertStringNotContainsString('Oolong', $html);
     }
+
+    /**
+     * @test
+     */
+    public function showActionTriggers404ForMissingTea(): void
```
oliverklee, line 84: Can we also test for the message?

oliverklee (changes requested):

Tests/Functional/Controller/TeaControllerTest.php
d-s-codappix: That's correct. I'll add this as another test.
I'd prefer to also have a test with an unavailable tea, as the code paths within TYPO3 might be different.

Classes/Controller/TeaController.php
```diff
@@ -24,9 +27,26 @@ public function indexAction(): ResponseInterface
...
+        if ($tea === null) {
+            $this->trigger404('No tea given.');
+        }
+
         $this->view->assign('tea', $tea);
         return $this->htmlResponse();
     }
+
+    /**
+     * Will throw exception to trigger 404.
```
oliverklee, line 41:
```suggestion
     * @throws PropagateResponseException
```

oliverklee (changes requested):

Tests/Unit/Controller/TeaControllerTest.php
d-s-codappix: Solved as the correct one already got merged.

Classes/Controller/TeaController.php
d-s-codappix: Done with a57ac7ab07250f53eb5688c20061f626d6ea288f.

Tests/Functional/Controller/TeaControllerTest.php
d-s-codappix: Done with a57ac7ab07250f53eb5688c20061f626d6ea288f.

Classes/Controller/TeaController.php
```diff
@@ -24,9 +27,31 @@ public function indexAction(): ResponseInterface
...
+        }
+
         $this->view->assign('tea', $tea);
         return $this->htmlResponse();
     }
+
+    /**
+     * @throws PropagateResponseException
+     *
+     * @throws PropagateResponseException
```
oliverklee, line 46: Now we have the annotation twice …

Tests/Functional/Controller/TeaControllerTest.php
```diff
@@ -101,4 +101,40 @@ public function showActionRendersTheGivenTeas(): void
...
+
+        $response = $this->executeFrontendSubRequest($request);
+
+        self::assertSame(404, $response->getStatusCode());
+    }
+
+    /**
+     * @test
+     */
+    public function showActionTriggering404ReasonIsRendered(): void
```
oliverklee, line 132:
```suggestion
    public function showActionFor404RendersReasonFor404(): void
```

oliverklee (changes requested):

Classes/Controller/TeaController.php
d-s-codappix: Yeah … GitHub Web UI … I thought so, fetched, but didn't see it twice. Looks like it is asynchronous. Will fix it.

d-s-codappix: Done with bce62b5430b80d74fea2d7460de6d03bde8a72d0.

Tests/Functional/Controller/TeaControllerTest.php
d-s-codappix: Done with Done with bce62b5430b80d74fea2d7460de6d03bde8a72d0.

oliverklee (approved):

PR #1645, opened 2025-04-07
Task/phpunit upgrade

linxpinx:
Remove --whitelist option and set source in xml


.github/workflows/ci.yml
```diff
@@ -267,9 +268,10 @@ jobs:
           - typo3-version: "^12.4"
             php-version: "8.1"
             composer-dependencies: lowest
-          - typo3-version: "^12.4"
-            php-version: "8.1"
-            composer-dependencies: highest
+# disabled due to issues with ParaTest
+#          - typo3-version: "^12.4"
+#            php-version: "8.1"
+#            composer-dependencies: highest
```
d-s-codappix, line 274: Wouldn't it make sense to not use paratest in that case, or use lower PHPUnit Versions that are compatible, instead of disabling checks?

d-s-codappix (approved):

.github/workflows/ci.yml
d-s-codappix: I'll remove paratest in a follow-up: https://github.com/TYPO3BestPractices/tea/issues/1647
Just as discussed in our call.

Issue #1647, opened 2025-04-07
Task: Remove paratest


d-s-codappix:
## Goal
We want to remove paratest.

## Acceptance Criteria
- [x] paratest is not used anymore.
- [x] The commented workflows from https://github.com/TYPO3BestPractices/tea/pull/1645/files are enabled again.
- [x] dependabot will update phpunit (below major version 11).

## Additional Information
This is a follow up to https://github.com/TYPO3BestPractices/tea/pull/1645 and related to https://github.com/TYPO3BestPractices/tea/issues/1574#issuecomment-2631328087 -> https://github.com/paratestphp/paratest/issues/890

PR #1648, opened 2025-04-07
[TASK] Remove brianium/paratest

d-s-codappix:
As this component is not necessary in our setup yet. It is also easy to setup, so no need for us to showcase this.

Also this adds some extra complexity to our setup due to: https://github.com/paratestphp/paratest/issues/890 which was the trigger to remove the component.

Resolves: #1647

### DIFF
--- a/.github/dependabot.yml
+++ b/.github/dependabot.yml
@@ -19,6 +19,8 @@ updates:
       - dependency-type: "development"
     ignore:
       - dependency-name: "doctrine/dbal"
+      - dependency-name: "phpunit/phpunit"
+        versions: [ "^11.0" ]
       - dependency-name: "symfony/console"
       - dependency-name: "symfony/translation"
       - dependency-name: "symfony/yaml"
--- a/.github/dependabot.yml
+++ b/.github/dependabot.yml
@@ -18,9 +18,7 @@ updates:
     allow:
       - dependency-type: "development"
     ignore:
-      - dependency-name: "brianium/paratest"
       - dependency-name: "doctrine/dbal"
-      - dependency-name: "phpunit/phpunit"
       - dependency-name: "symfony/console"
       - dependency-name: "symfony/translation"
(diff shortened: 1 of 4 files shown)


d-s-codappix:
@oliverklee I've no experience with dependabot. Can you verify that the goal to update phpunit/phpunit < 11 was reached?

.github/dependabot.yml
```diff
@@ -18,9 +18,7 @@ updates:
     allow:
       - dependency-type: "development"
     ignore:
-      - dependency-name: "brianium/paratest"
       - dependency-name: "doctrine/dbal"
-      - dependency-name: "phpunit/phpunit"
```
oliverklee, line 23: We still need this to avoid automatic upgrades to PHPUnit 11:

```yaml
- dependency-name: "phpunit/phpunit"
  versions: [ "^11.0" ]
```

composer.json
```diff
@@ -58,7 +57,7 @@
 		"phpstan/phpstan": "1.12.14",
 		"phpstan/phpstan-phpunit": "1.4.2",
 		"phpstan/phpstan-strict-rules": "1.6.1",
-		"phpunit/phpunit": "10.5.31 || 10.5.44",
+		"phpunit/phpunit": "^10.5",
```
oliverklee, line 60:
```suggestion
		"phpunit/phpunit": "10.5.45",
```

oliverklee (changes requested):

.github/dependabot.yml
d-s-codappix: Done with cb1e969490b5597f01393ce29cc6a87a191f0cee

composer.json
d-s-codappix: Done with cb1e969490b5597f01393ce29cc6a87a191f0cee

oliverklee (approved):

Issue #1575, opened 2025-02-03
Upgrade to TYPO3 testing framework 8.x


oliverklee:
## Goal

In preparation for TYPO 13LTS, we need a compatible version of the TYPO3 testing framework.

## Acceptance Criteria

- [ ] We now use `typo3/testing-framework` 8.2.7.
- [ ] All tests are still green.

## Additional Information

Blocked by #1574.

PR #1649, opened 2025-04-07
[TASK] Update typo3/testing-framework

d-s-codappix:
As preparation for TYPO3 v13 support.

Resolves: #1575

### DIFF
--- a/composer.json
+++ b/composer.json
@@ -72,7 +72,7 @@
 		"tomasvotruba/type-coverage": "1.0.0",
 		"typo3/cms-fluid-styled-content": "^12.4.26",
 		"typo3/coding-standards": "0.6.1 || 0.8.0",
-		"typo3/testing-framework": "^8.2.7",
+		"typo3/testing-framework": "8.2.7",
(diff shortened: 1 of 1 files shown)


composer.json
```diff
@@ -72,7 +72,7 @@
 		"tomasvotruba/type-coverage": "1.0.0",
 		"typo3/cms-fluid-styled-content": "^12.4.26",
 		"typo3/coding-standards": "0.6.1 || 0.8.0",
-		"typo3/testing-framework": "7.1.1",
+		"typo3/testing-framework": "^8.2.7",
```
oliverklee, line 75: We're using fixed versions here to avoid build failures due to upgraded dependency versions without any changes on our side:

```suggestion
		"typo3/testing-framework": "8.2.7",
```

oliverklee (changes requested):

composer.json
d-s-codappix: Done with f97be95558e243ba533bc56ae17e5a756e0cd097.

oliverklee (approved):

Issue #1570, opened 2025-02-03
Add flexforms for the Tea plugins


oliverklee:
## Goal

To increase usability, we want the Tea plugin to be configurable via Flexforms.

## Acceptance Criteria

- [ ] We have Flexforms configuration for `storagePid`, `recursion` and `singleViewPageUid` in the single view plugin.
- [ ] We have Flexforms configuration for `storagePid` and `recursion` in the single view plugin.
- [ ] TypoScript constants and setup for  `storagePid`, `recursion` and `singleViewPageUid` are gone.
- [ ] The existing functional controller tests test for these three settings.

oliverklee:
For the last part, we first need proper functional tests for creating teas in the FE editor.

oliverklee:
All done. Closing.

PR #1651, opened 2025-04-07
[FEATURE] Move the tea index configuration to FlexForms

oliverklee:
This is part 1.

(Part 2 is about migrating the storage page for new records created in the FE editor.)

Part of #1570

### DIFF
--- a/Configuration/FlexForms/TeaIndex.xml
+++ b/Configuration/FlexForms/TeaIndex.xml
@@ -1,0 +1,82 @@
+<?xml version="1.0" encoding="UTF-8"?>
+<T3DataStructure>
+    <sheets>
+        <pagesAndFolders>
+            <ROOT>
+                <TCEforms>
+                    <sheetTitle>
+                        LLL:EXT:tea/Resources/Private/Language/locallang.xlf:plugin.tea_index.settings.tab.pagesAndFolders
+                    </sheetTitle>
+                </TCEforms>
+                <type>array</type>
+                <el>
+                    <settings.singleViewPageUid>
+                        <TCEforms>
+                            <label>
+                                LLL:EXT:tea/Resources/Private/Language/locallang.xlf:plugin.tea_index.settings.pagesAndFolders.singleViewPageUid
+                            </label>
+                            <config>
+                                <type>group</type>
+                                <internal_type>db</internal_type>
+                                <allowed>pages</allowed>
+                                <size>1</size>
+                                <minitems>1</minitems>
+                                <maxitems>1</maxitems>
+                                <hideSuggest>1</hideSuggest>
+                            </config>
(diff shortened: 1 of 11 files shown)


Configuration/TCA/Overrides/tt_content.php
```diff
@@ -18,6 +19,13 @@ static function (): void {
             'EXT:tea/Resources/Public/Icons/Extension.svg'
         );
 
+        // These two commands add the flexform configuration for the plugin.
+        $GLOBALS['TCA']['tt_content']['types']['list']['subtypes_addlist']['tea_teaindex'] = 'pi_flexform';
+        ExtensionManagementUtility::addPiFlexFormValue(
+            'tea_teaindex',
```
DanielSiepmann, line 25: The string `tea_teaindex` should be our plugin signature.
This should be available as return value of above `registerPlugin()` call.
I guess it would make sense to refactor the lines to store the return value, e.g. in `$pluginSignature` and use it here:

```suggestion
        $GLOBALS['TCA']['tt_content']['types']['list']['subtypes_addlist'][$pluginSignature] = 'pi_flexform';
        ExtensionManagementUtility::addPiFlexFormValue(
            $pluginSignature,
```

This is a new feature since 12.0, and therefore available in our code base: https://docs.typo3.org/c/typo3/cms-core/main/en-us/Changelog/12.0/Feature-82809-MakeExtensionUtilityregisterPluginMethodReturnPluginSignature.html

Configuration/FlexForms/TeaIndex.xml
```diff
@@ -0,0 +1,82 @@
...
+                                <internal_type>db</internal_type>
+                                <allowed>pages</allowed>
+                                <size>1</size>
+                                <minitems>1</minitems>
+                                <maxitems>1</maxitems>
+                                <hideSuggest>1</hideSuggest>
+                            </config>
+                        </TCEforms>
+                    </settings.singleViewPageUid>
+                    <persistence.storagePid>
```
DanielSiepmann, line 29: Do we have a reason to build the two persistence options ourself, instead of enabling the `tt_content` fields?

DanielSiepmann (approved):

Configuration/FlexForms/TeaIndex.xml
DanielSiepmann: We discussed this, and will use this approach.
The core supports this, as well as two database fields.
But we prefer the FlexForm, as this bundles everything together in one file, and follows a single concept.

Configuration/TCA/Overrides/tt_content.php
oliverklee: Done.

DanielSiepmann (approved):

PR #1687, opened 2025-05-14
[TASK] Do not show the Composer version on CI anymore

oliverklee:
This will not be possible anymore when we switch CI to `runTests.sh`.

This is a pre-patch to switching unit and functional tests on GitHub Actions to `runTests.sh`.


kanow (approved):
Looks good to me.

Issue #1345, opened 2024-06-18
Drop the `-e` argument from `runTests.sh`


oliverklee:
## Goal
get rid of the `-e` solution because Argument passing using `--` works well.

## Acceptance criteria
- [x] the `-e`solution is removed

## Additional Information
- https://review.typo3.org/c/Packages/TYPO3.CMS/+/82728
- https://review.typo3.org/c/Packages/TYPO3.CMS/+/84745

oliverklee:
So we probably can apply/port https://review.typo3.org/c/Packages/TYPO3.CMS/+/84745 to remove the now-deprecated option.

oliverklee:
Reopening - this is not done yet.

PR #1689, opened 2025-05-18
[TASK] Unblock argument passing in `Build/Scripts/runTests.sh`

oliverklee:
The first argument after processing defined options has been stored into the variable `TEST_FILE`. This value was passed to PHPUnit based commands as last argument, defining the test case to execute (or a subset).

Removing this workaround and directly passing remaining options to the command makes special arguments like `-e` obsolete and unblocks more generic command passing.

This change removes this variable and passes all remaining arguments after option processing directly to the PHPUnit command and paves the way for future command improvements.

Note:   This makes the `-e` option basically obsolete, but
        is kept for now to minimize muscle memory issues.
        Using it will display a `deprecation` notice at the
        end of the script - so it gets recognized.

This is the Tea port of this change in the TYPO3 Core, but without test chunking, the deprecated unit or functional tests and without the acceptance tests (as we don't have any of those):

https://review.typo3.org/c/Packages/TYPO3.CMS/+/82728

This is the first part of #1345.

### DIFF
--- a/Build/Scripts/runTests.sh
+++ b/Build/Scripts/runTests.sh
@@ -239,12 +239,16 @@ Options:
             - 8.3: (default) use PHP 8.3
             - 8.4: use PHP 8.4
 
-    -e "<phpunit options>"
+    -e "<phpunit options>" (DEPRECATED).
         Only with -s functional|functionalDeprecated|unit|unitDeprecated|unitRandom
         Additional options to send to phpunit (unit & functional tests). For phpunit,
         options starting with "--" must be added after options starting with "-".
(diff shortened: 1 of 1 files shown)


oliverklee:
I'll still need to fix the shellcheck warnings.

oliverklee:
@cumuru If I remember correctly, you had this exact same warnings during our on-site sprint, and you found a way around it. Can you still remember how you solved this?

oliverklee (commented):
Looks good to me.

d-s-codappix (approved):

Issue #1569, opened 2025-02-03
Feature: Add functional Tests for FrontEndEditorController


DanielSiepmann:
## Goal
Provide concrete examples on how to test frontend controllers rendering data from the database involving frontend user.

## Acceptance Criteria
- [ ] Test covering the index action without logged in user
- [ ] Test covering the index action with logged in user but no teas
- [ ] Test covering the index action with logged in user and two teas

- [ ] Test covering the new action without logged in user
- [ ] Test covering the new action with logged in user

- [ ] Test covering the new action without logged in user
- [ ] Test covering the new action with logged in user and invalid tea
- [ ] Test covering the new action with logged in user and valid tea

- [ ] Test covering the edit action without logged in user but existing tea
- [ ] Test covering the edit action without logged in user and unavailable tea
- [ ] Test covering the edit action with logged in user but unavailable tea
- [ ] Test covering the edit action with logged in user and available tea

- [ ] Test covering the update action without logged in user but existing tea
- [ ] Test covering the update action without logged in user and unavailable tea
- [ ] Test covering the update action with logged in user but unavailable tea
- [ ] Test covering the update action with logged in user and available valid tea
- [ ] Test covering the update action with logged in user and available invalid tea

- [ ] Test covering the delete action without logged in user but existing tea
- [ ] Test covering the delete action without logged in user and unavailable tea
- [ ] Test covering the delete action with logged in user but unavailable tea
- [ ] Test covering the delete action with logged in user and available tea

## Additional Information
Anything else that would be helpful to know to implement this feature.

PR #1729, opened 2025-07-07
[TASK] Migrate controller tests to functional tests (indexAction)

DanielSiepmann:
It is considered best practice to cover controllers via functional instead of unit tests.
We therefore migrate the existing unit tests to functional tests.

The migration is split into multiple batches for smaller changes. This one covers tests related to `indexAction()`.

We can't add types to constants, as we still support PHP < 8.3.
We therefore completely disable this PHPStan issue for our project.

Relates: #1569


Tests/Functional/Controller/FrontEndEditorControllerTest.php
```diff
@@ -0,0 +1,50 @@
...
+    {
+        $request = (new InternalRequest())->withPageId(self::PAGE_UID);
+
+        $html = (string)$this->executeFrontendSubRequest($request)->getBody();
+
+        self::assertStringContainsString('Please configure this plugin to be only visible if a website user is logged in.', $html);
+    }
+
+    #[Test]
+    public function indexActionForLoggedInUserRendersTeasOwnedByTheLoggedInUser(): void
```
oliverklee, line 38: Please let's split this test into three tests:

- `indexActionForLoggedInUserRendersTeaOwnedByTheLoggedInUser`
- `indexActionForLoggedInUserDoesNotRenderTeaOwnedByOtherUser`
- `indexActionForLoggedInUserDoesNotRenderTeaWithoutOwner`

Tests/Functional/Controller/FrontEndEditorControllerTest.php
```diff
@@ -0,0 +1,50 @@
...
+    {
+        $this->importCSVDataSet(__DIR__ . '/Fixtures/Database/TeasAssignedToUser.csv');
```
oliverklee, line 40: Please let's move the  `indexAction`-specific tests into a subfolder:

`'/Fixtures/Database/indexAction/TeasAssignedToUser.csv');`

(I've made the experience that this save a lot of headaches when the number of tests for a controller grows.)

Tests/Functional/Controller/FrontEndEditorControllerTest.php
```diff
@@ -0,0 +1,50 @@
...
+    protected function setUp(): void
+    {
+        parent::setUp();
+
+        $this->importCSVDataSet(__DIR__ . '/Fixtures/Database/ContentElementTeaFrontEndEditor.csv');
+        $this->importCSVDataSet(__DIR__ . '/Fixtures/Database/FrontendUsers.csv');
+    }
+
+    #[Test]
+    public function indexActionForNoLoggedInUser(): void
```
oliverklee, line 28: Let's add a verb:

```suggestion
    public function indexActionForNoLoggedInUserRendersErrorMessage(): void
```

oliverklee (changes requested):

Tests/Functional/Controller/FrontEndEditorControllerTest.php
oliverklee: Absolutely! (I actually already do this in the tests of my extension and forgot this in my comment.)

Tests/Functional/Controller/Fixtures/Database/FrontEndEditorController/TeasAssignedToUser.csv
```diff
@@ -0,0 +1,5 @@
+"tx_tea_domain_model_tea"
+,"uid","pid","title","owner"
+,1,2,"Godesberger Burgtee",1
+,2,2,"Oolong",2
+,3,2,"Sencha",0
```
oliverklee, line 5: Now that we have 3 separate tests for the different cases, we should also have 3 different fixtures:

- `TeaOwnedByLoggedInUser`
- `TeaOwnedByOtherUser`
- `TeaWithoutOwner`

Tests/Functional/Controller/FrontEndEditorControllerTest.php
```diff
@@ -0,0 +1,73 @@
...
+
+        $html = $this->getHtml();
+
+        self::assertStringNotContainsString('Sencha', $html);
+    }
+
+    private function getHtml(): string
+    {
+        $request = (new InternalRequest())->withPageId(self::PAGE_UID);
+        $context = (new InternalRequestContext())->withFrontendUserId(1);
```
oliverklee, line 69: Let's have a constant for the FE user UID.

Tests/Functional/Controller/FrontEndEditorControllerTest.php
```diff
@@ -0,0 +1,73 @@
...
+    public function indexActionForLoggedInUserDoesNotRenderTeaWithoutOwner(): void
+    {
+        $this->importCSVDataSet(__DIR__ . '/Fixtures/Database/FrontEndEditorController/TeasAssignedToUser.csv');
+
+        $html = $this->getHtml();
+
+        self::assertStringNotContainsString('Sencha', $html);
+    }
+
+    private function getHtml(): string
```
oliverklee, line 66:
```suggestion
    private function getHtmlWithLoggedInUser(): string
```

oliverklee (changes requested):

Tests/Functional/Controller/Fixtures/Database/FrontEndEditorController/TeasAssignedToUser.csv
DanielSiepmann: I'm not a fan of that, as that would split the UIDs into three different files, which would make it hard to have an overview of the available UIDs and what they mean.

Tests/Functional/Controller/FrontEndEditorControllerTest.php
DanielSiepmann: I'm not a fan to introduce another hop while reading code for a single location, which also already communicates the actual meaning of the hard coded value by the "word" in front of it `FrontendUserId`

Tests/Functional/Controller/Fixtures/Database/FrontEndEditorController/FrontendUsers.csv
```diff
@@ -0,0 +1,3 @@
+"fe_users"
+,"uid","pid","username","password"
```
oliverklee, line 2: We can drop the `password` column from this fixture file as it's not relevant for these tests.

oliverklee (changes requested):

oliverklee (approved):

Issue #1587, opened 2025-02-03
Task: Please add a speaking title


kanow:
## Goal
I as a user/developer has all informations how  to execute tests

## Acceptance Criteria
- [ ]  explain how to execute test with docker (not in the local system)
- [ ]  explain how to execute functional tests with a database

## Additional information
I start to execute some tests and had errors. I'm searching for information to execute tests in the documentation but the infor not detailed enough.

kanow:
I looked over the documentation in `Running.rst` file and I would ay that the first acceptance criteria is also solved bei the text. Ist that ok for you?

kanow:
Ok, then I close this issue.

PR #1737, opened 2025-07-07
[DOCS] Using `runTests.sh` also for composer commands

kanow:
Related: #1587

### DIFF
--- a/Documentation/Running.rst
+++ b/Documentation/Running.rst
@@ -31,7 +31,9 @@ the `composer.json`.
 If you have problems with missing dependencies on your local machine, it is
 recommended to execute tests with the usage of the `runTests.sh` script.
 
-Example: `./Build/Scripts/runTests.sh -s composer ci:php:lint`
+.. code-block:: bash
+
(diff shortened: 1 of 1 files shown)


Documentation/Running.rst
```diff
@@ -53,91 +55,91 @@ You can currently run these code checks on the command line:
...
-    composer ci:json:lint
+    ./Build/Scripts/runTests.sh -s composer ci:json:lint
 
 Lints the JSON files.
 
 .. index:: Commands; composer ci:php
 .. code-block:: bash
 
-    composer ci:php
+ ./Build/Scripts/runTests.sh -s composer ci:php
```
oliverklee, line 72: The indentation for code blocks here seems inconsistent to me: Here, it's 2 spaces, while above, it's 4 spaces.

This should be consistent in all changed lines.

The current recommendation is 4 spaces AFAIK (at least that's what the later core docs use).

oliverklee (changes requested):

oliverklee (approved):

PR #1767, opened 2025-08-04
1332 make plugin a ctype

kanow:
Started refactoring to use CTypes for plugins


linxpinx:
@oliverklee ah ok but the wizard works. I test it in a second TYPO3 Installation and the Wizard change from list_type to CType and the Plugin works after running the wizard. Is it possible that the Class tooks the interface from the Installation TYPO3 if there is no Interface in our Test TYPO3?

linxpinx:
This PR could be closed when #1811 is done.

PR #1768, opened 2025-08-04
Allow ssch/typo3 rector v3 and phpstan/phpstan v2

linxpinx:
Switching between the Versions TYPO3 12.4 13.4 with the right Versions for ssch/typo3 and phpstan/phpstan including the dependencies with the right Versions works. 
Maybe you had to clean up the System, when you switch with composerUpdateMax to composerUpdateMin with different TYPO3 Versions.

**Examples:**
Updates to TYPO3 V13.4 maximum
`./Build/Scripts/runTests.sh -s composerUpdateMax -t 13.4`
Updates to TYPO3 V12.4 maximum
`./Build/Scripts/runTests.sh -s composerUpdateMax -t 12.4`
Updates to TYPO3 V13.4 minimum
`./Build/Scripts/runTests.sh -s composerUpdateMin -t 13.4`
Updates to TYPO3 V12.4 maximum
`./Build/Scripts/runTests.sh -s composerUpdateMin -t 12.4`
Updates the current TYPO3 Version to maximum
`./Build/Scripts/runTests.sh -s composer update`


composer.json
```diff
@@ -41,35 +41,36 @@
...
-		"phpstan/phpstan": "1.12.27",
-		"phpstan/phpstan-phpunit": "1.4.2",
-		"phpstan/phpstan-strict-rules": "1.6.2",
+		"phpstan/phpstan": "1.12.27 || ^2.1.19",
+		"phpstan/phpstan-phpunit": "1.4.2 || ^2.0.7",
+		"phpstan/phpstan-strict-rules": "1.6.2 || ^2.0.6",
 		"phpunit/phpunit": "10.5.47",
-		"rector/type-perfect": "1.0.0",
-		"saschaegerer/phpstan-typo3": "1.10.2",
+		"rector/type-perfect": "1.0.0 || ^2.1.0",
```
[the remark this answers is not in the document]
linxpinx: ok i checked against composer.json from main branch and changed the constraints

composer.json
```diff
@@ -105,7 +106,10 @@
 			"*": "dist"
 		},
 		"sort-packages": true,
-		"vendor-dir": ".Build/vendor"
+		"vendor-dir": ".Build/vendor",
+		"platform": {
+			"php": "8.3.0"
```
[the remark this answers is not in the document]
linxpinx: yes I comes from composerUpdateMin. I removed it.

composer.json
```diff
@@ -41,35 +41,35 @@
...
-		"symfony/translation": "6.4.24 || 7.3.3",
-		"symfony/yaml": "6.4.25 || 7.3.3",
-		"tomasvotruba/cognitive-complexity": "0.2.3",
-		"tomasvotruba/type-coverage": "1.0.0",
-		"typo3/cms-fluid-styled-content": "^12.4.31",
+		"ssch/typo3-rector": "2.15.2 || 3.1.1",
+		"ssch/typo3-rector-testing-framework": "2.0.1 || 3.0.0",
+		"symfony/console": "6.4.25 || 7.3.1",
+		"symfony/translation": "6.4.24 || 7.3.1",
+		"symfony/yaml": "6.4.25 || 7.3.1",
```
oliverklee, line 69: We have update the development dependencies (including the Symfony development dependencies) in #1781. So these all should have 7.3.3 as highest version, not 7.3.1 anymore.

oliverklee (changes requested):
It looks like we also need to update the PHPStan configuration and drop/change removed configuration options.

oliverklee (changes requested):
Also, we should rebase on `main` (instead of merging `main` into this branch) to get rid of the corresponding merge commit.

Tests/Functional/Command/CreateTestDataCommandTest.php
```diff
@@ -41,6 +41,9 @@ protected function setUp(): void
     #[Test]
     public function isConsoleCommand(): void
     {
+        // The CreateTestDataCommand extends Command, so this test is always true
+        // This test validates the inheritance structure is correct
+        /** @phpstan-ignore-next-line staticMethod.alreadyNarrowedType */
```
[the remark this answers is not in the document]
oliverklee: The configuration entry for disabling a rule (or ignoring certain warnings) for the folder `Tests/` looks like this:

```
  ignoreErrors:
    -
      message: '#^Call to static method PHPUnit\\Framework\\Assert\:\:assertInstanceOf\(\) .* will always evaluate to#'
      path: '../../Tests/'
```

linxpinx:
> Also, we should rebase on `main` (instead of merging `main` into this branch) to get rid of the corresponding merge commit.

I did a rebase in Fork when I take the changes from the main branch. Maybe you explain/show me how it had to be

Tests/Functional/Command/CreateTestDataCommandTest.php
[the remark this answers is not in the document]
linxpinx: s.o.

composer.json
```diff
@@ -41,35 +41,35 @@
...
 		"helmich/typo3-typoscript-lint": "^3.3.0",
-		"icanhazstring/composer-unused": "0.8.11",
+		"icanhazstring/composer-unused": "0.8.11 || 0.9.3",
 		"php-parallel-lint/php-parallel-lint": "1.4.0",
 		"phpmd/phpmd": "2.15.0",
 		"phpstan/extension-installer": "1.4.3",
-		"phpstan/phpstan": "1.12.28",
-		"phpstan/phpstan-phpunit": "1.4.2",
-		"phpstan/phpstan-strict-rules": "1.6.2",
+		"phpstan/phpstan": "1.12.28 || 2.1.19",
```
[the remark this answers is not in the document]
linxpinx: updated

Configuration/TCA/Overrides/tt_content.php
```diff
@@ -42,8 +43,11 @@ static function (): void {
 
         // This removes the default controls from the plugins.
         $controlsToRemove = 'recursive,pages';
+        /** @phpstan-ignore-next-line offsetAccess.nonOffsetAccessible */
         $GLOBALS['TCA']['tt_content']['types']['list']['subtypes_excludelist'][$indexPluginSignature] = $controlsToRemove;
+        /** @phpstan-ignore-next-line offsetAccess.nonOffsetAccessible */
         $GLOBALS['TCA']['tt_content']['types']['list']['subtypes_excludelist'][$showPluginSignature] = $controlsToRemove;
+        /** @phpstan-ignore-next-line offsetAccess.nonOffsetAccessible */
```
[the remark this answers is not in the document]
oliverklee: I was referring to the baseline (which we autogenerate using `composer phpstan:baseline`), not the `phpstan.neon` configuration file.

Generally, we follow this approach:

- comments in the code for single occurrences that we cannot avoid with "real code" (assertions, asserts), and than we don't want to ignore in general (e.g., passing invalid values to methods to test for exceptions)
- configuration file ignores for errors that we generally don't agree with (e.g., instanceof checks in the tests)
- baseline for warnings that are valid and that we want to fix some day

Build/phpstan/phpstan.neon
```diff
@@ -20,8 +20,8 @@ parameters:
     - ../../Tests
     - ../../ext_localconf.php
 
-  # Allow instanceof checks, particularly in tests
-  checkAlwaysTrueCheckTypeFunctionCall: false
+  # Don't report errors for ignored patterns that don't match (version-specific ignores)
+  reportUnmatchedIgnoredErrors: false
```
oliverklee, line 24: Why this change?

(If I understand this option correctly, this helps us keep the baseline up to date and the configuration small.)

linxpinx: I use this options, cause we supported 2 TYPO3 Versions and for example: 
without this option in TYPO3 12.4 you will get
`Ignored error pattern #Cannot access offset .+ on mixed# in path /Users/admin_bernd/Development/best-practice-team/t312/vendor/ttn/tea/Configuration/TCA/Overrides/* was not matched in reported errors. ` while I TYPO3 13.4 phpstan is fine. Do you have a better solution for that issue?

oliverklee: We're running PHPStan with 1 TYPO3 version only (the higher version, i.e., 13LTS with this PR). So it's okay if the results in lower versions are different - and we don't need to cater for that.

linxpinx: @oliverklee I changed phpstan configuration. Maybe it's a good idea to but this information into the Documentation, cause I as a user would assume that the composer scripts run in both TYPO3 versions. Obviously, I assumed that ;-).

Build/phpstan/phpstan-baseline.neon
```diff
@@ -1,2 +1,5 @@
 parameters:
-	ignoreErrors: []
+	ignoreErrors:
+		-
+			message: '#^Call to static method PHPUnit\\Framework\\Assert\:\:assertInstanceOf\(\) .* will always evaluate to#'
```
oliverklee, line 4: This configuration is for something that we want to be excluded. So it should go into `phpstan.neon`, not into the baseline.

The baseline must not contain manual entries as it's expected to be generated by `composer phpstan:baseline`, and manual entries would get overwritten the next time someone regenerates the baseline.

Classes/Domain/Repository/TeaRepository.php
```diff
@@ -21,6 +21,7 @@ class TeaRepository extends Repository
      *
      * @return QueryResultInterface<Tea>
      */
+    /** @phpstan-ignore-next-line missingType.generics */
```
oliverklee, line 24: This is something that we want to fix (in the Core, in our code, or with changing a type annotation). So we shouldn't have this blocklist line.

Tests/Functional/Command/CreateTestDataCommandTest.php
```diff
@@ -41,6 +41,8 @@ protected function setUp(): void
     #[Test]
     public function isConsoleCommand(): void
     {
+        // The CreateTestDataCommand extends Command, so this test is always true
+        // This test validates the inheritance structure is correct
```
oliverklee, line 45: I don't think we should have a command above each and every `assertInstanceOf` in the tests. The fact that we have that call implicitly communicates that we want to check the inheritance chain.

Ditto for the other added commands above the other `assertInstanceOf` calls.

oliverklee (changes requested):
This branch also needs a rebase.

Tests/Unit/Domain/Model/TeaTest.php
```diff
@@ -26,6 +26,7 @@ protected function setUp(): void
     #[Test]
     public function isAbstractEntity(): void
     {
+        // Tea extends AbstractEntity, so this test validates the inheritance structure
```
oliverklee, line 29: This comment can also be removed.

oliverklee (changes requested):
Looks very good already. Just one more comment to remove.

PR #1778, opened 2025-09-01
[TASK] Migrate controller tests to functional tests (newAction)

d-s-codappix:
It is considered best practice to cover controllers via functional instead of unit tests.
We therefore migrate the existing unit tests to functional tests.

The migration is split into multiple batches for smaller changes. This one covers tests related to `newAction()`.

We remove the `$tea` argument from `newAction()`, as the extension does not provide a way in itself to call this action with this argument.

Relates: #1569

### DIFF
--- a/Tests/Functional/Controller/FrontEndEditorControllerTest.php
+++ b/Tests/Functional/Controller/FrontEndEditorControllerTest.php
@@ -201,6 +201,16 @@ final class FrontEndEditorControllerTest extends AbstractFrontendControllerTestC
         ]);
     }
 
+    #[Test]
+    public function newActionCanBeCalledByUser(): void
+    {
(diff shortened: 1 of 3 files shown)


Tests/Functional/Controller/FrontEndEditorControllerTest.php
```diff
@@ -201,6 +201,16 @@ public function updateActionWithTeaWithoutOwnerThrowsException(): void
         ]);
     }
 
+    #[Test]
+    public function newActionCanBeCalledByUser(): void
```
oliverklee, line 205:
```suggestion
    public function newActionHasLabelForNewAction(): void
```

oliverklee (changes requested):

oliverklee (approved):

PR #1783, opened 2025-09-01
[TASK] Move "composer unused" configuration from root to Build folder

d-s-codappix:
The project root should be as clean as possible.
Configuration files for tools should be within a dedicated folder like `Build/`.
That's why we also move the configuration file for composer-unused.

This has the drawback that the tool can't be used with defaults, but explicitly needs to be called with `--configuration` option.

Resolves: #1782


d-s-codappix:
> We also should remove the corresponding entry from `.gitattributes`.

Thanks, good point, my full text search didn't search that file.
PR is updated.

PR #1814, opened 2025-09-25
[FEATURE] Add support for PHP 8.5

oliverklee:


oliverklee:
Note: The code coverage job breakage is unrelated - https://coveralls.io/ seems to error out at the moment.

PR #1815, opened 2025-09-25
[TASK] Re-enable a skipped test on CI

oliverklee:
This tests failed in the past and now is green.

So let's re-enable it.


oliverklee:
Note: The code coverage job breakage is unrelated - https://coveralls.io/ seems to error out at the moment.

PR #1824, opened 2025-10-15
[TASK] Update the PHPStan baseline

### LINKED ISSUE #1821: PHPStan build is failing on CI
https://github.com/TYPO3BestPractices/tea/actions/runs/18462694847/job/52597592001

oliverklee:
Fixes #1821


oliverklee:
> But I'd appreciate it if we create a follow up to solve the issues instead of adding them to the baseline.

Yes, I'll take care of that.

PR #1826, opened 2025-10-16
[TASK] Switch PHPStan back to TYPO3 V12 on CI

oliverklee:
At the moment, we have TYPO3 12LTS as the default version in `runTest.sh`, and PHPStan should show now warnings if it run with the default settings. This makes the life of developers easier.


d-s-codappix:
Approved and merged for now. But we should run QA for all combinations of supported PHP + TYPO3 versions via matrix, shouldn't we?

EDIT: Just like done for tests.

PR #1837, opened 2025-10-30
[TASK] Make the Dependabot blocklist more future-proof

oliverklee:
Use an unlimited version range instead of specific major versions.


linxpinx (approved):
ok

Issue #1751, opened 2025-07-21
Task: Migrate composer script `ci:composer:normalize` to `runTests.sh`


DanielSiepmann:
## Goal
Get rid of existing composer script `ci:composer:normalize`.
See: https://github.com/TYPO3BestPractices/tea/issues/1750

## Acceptance Criteria
- [x] ~~Remove from `composer.json`~~
- [x] Implement in `runTests.sh` (if necessary)
- [x] Migrate GitHub Actions (if necessary)
- [x] ~~Migrate GitLab Actions (if necessary)~~

## Additional Information
Anything else that would be helpful to know to complete this task.

PR #1851, opened 2025-11-03
[TASK] Migrate composer normalize in gitlab action

kanow:
Change gitlab action to use `runTests.sh` for composer normalize.

Related: #1751


d-s-codappix:
Fine for me. But I guess the target is not to use `runTest.sh` as another layer/wrapper, but instead of the composer task.
So get rid of the composer script and use runtest instead.
What do you say @oliverklee?

Issue #1846, opened 2025-11-03
Port more functional tests for the FE editor from seminars to Tea


oliverklee:
## Goal

As a developer, I want to cover more business-relevant functionality for the FE editor so I can be sure that the code behaves as intended and that changes won't break the functionality.

## Acceptance Criteria

- [ ] We have a list of functional tests for the FE editor [the seminars extension has](https://github.com/oliverklee/ext-seminars/blob/main/Tests/Functional/Controller/FrontEndEditorControllerTest.php) (that we don't have yet) as additional checkboxes in this issue.
- [ ] We have those tests as well (possibly in multiple PRs to make reviews quicker).

PR #1854, opened 2025-11-03
[TASK] Add more tests for FrontEndEditorController

d-s-codappix:
To cover more functionalities of the controller and template.

Resolves: #1846

### DIFF
--- a/Tests/Functional/Controller/FrontEndEditorControllerTest.php
+++ b/Tests/Functional/Controller/FrontEndEditorControllerTest.php
@@ -82,8 +82,8 @@ final class FrontEndEditorControllerTest extends AbstractFrontendControllerTestC
 
         $html = $this->getHtmlWithLoggedInUser();
 
-        $expected = LocalizationUtility::translate('plugin.frontEndEditor.index.heading', 'tea')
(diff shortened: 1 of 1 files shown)


Tests/Functional/Controller/FrontEndEditorControllerTest.php
```diff
@@ -71,6 +73,70 @@ public function indexActionForLoggedInUserDoesNotRenderTeaWithoutOwner(): void
...
 
+    #[Test]
+    public function indexActionHasHeadline(): void
+    {
+        $this->importCSVDataSet(__DIR__ . '/Fixtures/Database/FrontEndEditorController/TeaAssignedToLoggedInUser.csv');
+
+        $html = $this->getHtmlWithLoggedInUser();
+
+        $expected = LocalizationUtility::translate('plugin.frontEndEditor.index.heading', 'tea')
+            ?? throw new \RuntimeException('Could not fetch translation value.', 1762183380);
```
oliverklee: I'd prefer to not use the `??` operator to trigger method calls.

In this particular case, I'd say we add an `assertIsString` so that PHPUnit will let us know if the translation cannot be fetched.

oliverklee (approved):
LGTM.

PR #1930, opened 2026-01-22
[DOCS] Drop the download statistics badge from the README

oliverklee:
This vanity badge does not really provide helpful information.


oliverklee:
Thanks! I've put this up for decision for our next team call.

PR #1936, opened 2026-01-24
[BUGFIX] Return exitcode != 0 when `runTests.sh -s composerUpdate(Min|Max)` fails

### LINKED ISSUE #1882: `composerUpdateMax` and `composerUpdateMin` should return an error exit code if installation fails
## Short problem description

When `composerUpdate(Min|Max)` from our `runTests.sh` fails (e.g., due to a repository not being reachable or the requirements creating unsolvable dependency problems), the process still returns a success return code. Then the next CI steps get executed and runs into problems because the Composer packages are not installed.

1. create unsolvable dependencies in `composer.json`, e.g. `"php": "^5.6"`.
2. run `./Build/Scripts/runTests.sh -s composerUpdateMax`
3. `echo $?` to show the exit code

## actual behavior

Exit code 0

Also, `composerUpdateMax` continues to show the installed packages etc.

## expected behavior

Exit code 1

Also, `composerUpdateMax` exits after the first problem as the next steps (show installed packages etc.) do not make sense anymore.

## Additional Information

We also should port this fix to the TYPO3 Core.

[Info on exit codes](https://tldp.org/LDP/abs/html/exitcodes.html)

sbuerk:
The two `Build/Scripts/runTests.sh -s composerUpdateMin` and
`Build/Scripts/runTests.sh -s composerUpdateMax` uses multiple
commands passed to a single php container as chained command
concatenated using `;` executing all commands, not stopping
if one failes only returning the exit code of the last command
`composer show`, which is mostly `0` and communication success.

This change reworks the command building using `&&` to chain
the commands and abort in case one fails and using exitcode
of the failed command.

Resolves: #1882

### DIFF
--- a/Build/Scripts/runTests.sh
+++ b/Build/Scripts/runTests.sh
@@ -568,14 +568,14 @@ case ${TEST_SUITE} in
         ;;
     composerUpdateMax)
         # `dumpautoload` removed due to error with missing `composer.lock` file on publishing public assets.
-        COMMAND="composer config --unset platform.php; composer require --no-ansi --no-interaction --no-progress --no-install typo3/minimal:"^${CORE_VERSION}"; composer update --no-progress --no-interaction; composer show"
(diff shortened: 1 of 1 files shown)


oliverklee:
@sbuerk We've updated a few things in the CI configuration and in `runTest.sh`. Could you please rebase? Thanks!

oliverklee (approved):
Thanks a lot! ❤️

PR #1962, opened 2026-02-02
[TASK] Drop redundant assignment in `runTests.sh`

### LINKED ISSUE #1959: Drop redundant assignment in `runTests.sh`
## Story

As a developer, I want my code to not have redundant statements so that the code is easier to understand.

## Acceptance criteria

- [ ] In the `composerNormalize` command in `runTests.sh`, the first `COMMAND=` (above the `if`) is removed.

oliverklee:
The assign variable is overwritten in all branches of the following condition.

Fixes #1959


kanow (approved):
Yes, that make sanse.

PR #1969, opened 2026-02-22
[TASK] Switch coverage collection to `runTests.sh` on `GitHub`

### LINKED ISSUE #1802: Switch the code coverage collection on GitHubActions to use `runTests.sh`

sbuerk:
This replaces the original content of this pull request. The branch is rebased
onto current `main` and the work was redone, because the first approach did not
hold up: it kept installing PHP, Xdebug and the dependencies on the runner, and
it was built on the assumption that the report path is fixed by the Coveralls
action, which turned out not to be true.

The series is split into five commits which build on each other and can be
extracted into separate pull requests with their own issues. Only the fourth
resolves #1802, the others carry it as `Related`.

## The commits

1. **Write coverage artefacts to `.Build/`** — the reports were the last thing
   here using a lowercase `build/` directory, which collides with the tracked
   `Build/` on case-insensitive filesystems. `.Build/` is already ignored
   through `/.Build/*` and already removed by `prepare-release`, so the
   artefacts can neither collide with a tracked directory nor be committed by
   accident, and the `/build` ignore entry is dropped rather than replaced.
2. **Pass the Coveralls action its inputs** — `github-token` and `file` were
   passed in an `env:` block. Both are action *inputs*, so both were discarded:
   the action ran as `coveralls report --no-fail` and only found the report
   through its auto-detection, which happens to look for a clover file in a few
   well-known places. This is a real bug on `main` today and is independent of
   everything else in this series.
3. **Collect code coverage in `runTests.sh`** — new `-m` option for the `unit`,
   `unitRandom` and `functional` suites, plus a `coverageMerge` suite.
4. **Run the coverage workflow via `runTests.sh`** — what #1802 asks for. The
   workflow drops `setup-php`, the manual Composer steps and the system MySQL,
   and calls the harness like every other workflow. It also fixes the Composer
   cache path, which pointed at `~/.cache/composer` while the harness uses
   `.cache/composer`, so the cache never held anything.
5. **Drop the superseded coverage Composer scripts** — `check:coverage*` and
   `coverage:create-directories` bypassed the container environment and could
   not select the TYPO3 version, so they excluded the version-pinned tests of
   *both* versions at once and never covered what the suites actually run.

## Two decisions worth reviewing

**Coverage is not fanned out across the CI matrix.** The obvious design is one
report per matrix leg (8 unit + 14 functional), uploaded as artefacts and merged
at the end. It was worked out in full and then rejected.

A `.cov` file is a bare `serialize()` of the coverage object and carries no
version information up to php-code-coverage 12.5. `phpcov merge` — 10.0.1 being
the only line supporting PHPUnit 11.5 on PHP 8.2 — `include`s each file and
calls `merge()` with only an `instanceof` check. Upstream considered this risky
enough to add `assertPhpVersionMatches` and `assertDriverMatches` in
php-code-coverage 14, which needs PHP 8.4 and is out of reach here.

Since `typo3/testing-framework` 9.6.1 allows `phpunit: ^11.2.5 || ^12.1.2 ||
^13.0.2`, the 13.4 and 14.3 dependency sets can resolve different PHPUnit and
php-code-coverage versions. Merging across them would silently produce a
plausible but wrong number, and a wrong coverage figure is worse than none.

Because nothing empties `.Build/coverage/` on its own, the same trap exists
locally: a report from an earlier run with a different `-t` would still be
lying around. `coverageMerge` therefore refuses to merge unless every collected
report belongs to the same TYPO3 and PHP version, and `-s cleanTests` removes
the collected reports.

For what it is worth, `phpunit/phpunit` itself does the same thing: coverage
from exactly one job and one PHP version, with its large test matrices running
at `coverage: none`.

**Xdebug, because there is no alternative.** The `core-testing` images ship
Xdebug and not pcov, so `-m` runs the container with `XDEBUG_MODE=coverage`.
That is also why `-m` and `-x` cannot be combined, and the script says so
instead of collecting nothing silently.

## Verified locally

* `-s unit -m` and `-s functional -m` each write a report named after the TYPO3
  version, the PHP version and the DBMS; `-s coverageMerge` merges them into
  `.Build/logs/clover.xml` covering all six classes.
* A run without `-m` creates no `.Build/coverage/` at all.
* Placing a report of another TYPO3 version next to the collected ones makes
  `coverageMerge` refuse with an explanation instead of merging them.
* `-m` with `-x`, `-m` on a suite which runs no tests, and merging with nothing
  collected all exit 1 with an explanation.
* Green: `composer validate`, `lintJson`, `lintYaml`, `composerNormalize -n`,
  `composerUnused`, `psrVerify`, `shellcheck`, `unit`, `functional`.

## Note on coverage for pull requests from forks

Worth recording, since this pull request comes from a fork.

The upload itself works from a fork: this branch produced Coveralls build
79217076 at 96.64%, and job 181127924 before that. The build is keyed to the
synthetic merge commit GitHub creates for the pull request, which is why
looking it up by the head SHA returns nothing.

What never arrived on this pull request is a Coveralls comment or check.
Pull requests from branches in the repository do get the comment. Coveralls
posts no commit *status* anywhere in this repository, on `main` included, so a
Coveralls status could not be made a required check even if someone wanted to —
and it should not be, because it would never be delivered for a pull request
from a fork and would block every external contribution.

The previous head of this branch dates from February and predates ten of the
status checks the `main-ci` ruleset now requires. Those are produced by the
matrix this rebase brings in.

### DIFF
--- a/.github/dependabot.yml
+++ b/.github/dependabot.yml
@@ -25,6 +25,10 @@ updates:
       - dependency-name: "phpstan/phpstan-strict-rules"
       - dependency-name: "phpunit/phpunit"
         versions: [ ">= 11.0" ]
+      # pcov for phpunit 10, version match is currently one version off. Check packagist
+      # requirements again raising phpunit/phpunit version and align it correspondingly.
+      - dependency-name: "phpunit/pcov"
+        versions: [ ">= 10.0" ]
       - dependency-name: "rector/type-perfect"
       - dependency-name: "saschaegerer/phpstan-typo3"
       - dependency-name: "ssch/typo3-*"
--- a/.github/workflows/codecoverage.yml
+++ b/.github/workflows/codecoverage.yml
@@ -9,83 +9,57 @@ on:
 permissions:
   contents: read
 jobs:
+  # @todo Rethink the whole implementation. It's possible to consider creating the part coverage logs directly
+  #       for all unit, unitRandom and functional tests in the `ci.yml` workflow and using this only to merge
+  #       and upload it. That would avoid reexecuting only for a subset of matrix run here again.
+  #       In general there are two different approaches when doing the above mentioned considerations:
+  #
+  #       * Use job artifacts to get all of the `Build/coverage/*.pcov` files and merge it using the pcov tool
+  #       * Use parallel partly upload in ci.yaml unit, unitRandom, functional test runs and use this to finish
+  #         the parallel execution. See: https://github.com/coverallsapp/github-action#complete-parallel-job-example
   code-coverage:
     name: Calculate code coverage
     permissions:
       actions: write
     runs-on: ubuntu-24.04
-    env:
-      DB_DATABASE: typo3
-      DB_USER: root
-      DB_PASSWORD: root
-      DB_HOST: localhost
     steps:
       - name: Checkout
         uses: actions/checkout@v6
-      - name: Install PHP
-        uses: shivammathur/setup-php@v2
-        with:
-          php-version: "${{ matrix.php-version }}"
-          ini-file: development
-          tools: composer:v2, phive
-          extensions: mysqli
-          coverage: xdebug
-      - name: Install development tools
-        run: phive --no-progress install --trust-gpg-keys D8406D0D82947747293778314AA394086372C20A
-      - name: Show Composer version
-        run: composer --version
+      - name: Show the Composer version
+        run: ./Build/Scripts/runTests.sh -s composer -- --version
       - name: Show the Composer configuration
-        run: composer config --global --list
-      - name: Cache dependencies installed with composer
-        uses: actions/cache@v5
-        with:
-          key: "php-${{ matrix.php-version }}-typo3-${{ matrix.typo3-version }}-${{ matrix.composer-dependencies }}-composer-${{ hashFiles('**/composer.json') }}"
-          path: ~/.cache/composer
-          restore-keys: "php-${{ matrix.php-version }}-typo3-${{ matrix.typo3-version }}-${{ matrix.composer-dependencies }}-composer-\n"
-      - name: Install TYPO3 Core
-        env:
-          TYPO3: "${{ matrix.typo3-version }}"
+        run:  ./Build/Scripts/runTests.sh -s composer config --global --list
+      - name: Install composer dependencies
         run: |
-          composer require --no-ansi --no-interaction --no-progress --no-install typo3/cms-core:^"$TYPO3"
-          composer show
-      - name: Install lowest dependencies with composer
-        if: "matrix.composer-dependencies == 'Min'"
-        run: |
-          composer update --no-ansi --no-interaction --no-progress --with-dependencies --prefer-lowest
-          composer show
-      - name: Install highest dependencies with composer
-        if: "matrix.composer-dependencies == 'Max'"
-        run: |
-          composer update --no-ansi --no-interaction --no-progress --with-dependencies
-          composer show
-      - name: Start MySQL
-        run: "sudo /etc/init.d/mysql start"
+          ./Build/Scripts/runTests.sh -p ${{ matrix.php-version }} -t ${{matrix.typo3-version}} -s composerUpdate${{matrix.composer-dependencies}}
+      # @todo Consider to execute all unit tests with coverage and using job artifacts to merge them in one place
+      #       instead of executing only one run here with coverage. Could also be combined directly uploading the
+      #       parallel executions as partly upload and using this workflow only to make the final webhook call.
       - name: Run unit tests with coverage
-        run: composer check:coverage:unit
-      - name: Show generated coverage files
-        run: "ls -lahR build/coverage/"
+        run: |
+          ./Build/Scripts/runTests.sh -p ${{ matrix.php-version }} -s unit -m
+      # @todo Consider to execute all functional tests with coverage and using job artifacts to merge them in one place
+      #       instead of executing only one run here with coverage. Could also be combined directly uploading the
+      #       parallel executions as partly upload and using this workflow only to make the final webhook call.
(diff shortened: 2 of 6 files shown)


oliverklee:
@sbuerk We've updated a few things in the CI configuration and in `runTest.sh`. Could you please rebase? Thanks!

sbuerk:
Coveralls gets the call and the file, I don't get why it is not reported back as PR comment.

See the check pipeline: https://github.com/TYPO3BestPractices/tea/actions/runs/25213017521/job/73927396984?pr=1969

```
 
⠀⠀⠀⠀⠀⠀⣿
⠀⠀⠀⠀⠀⣼⣿⣧⠀⠀⠀⠀⠀⠀⠀ ⣠⣶⣾⣿⡇⢀⣴⣾⣿⣷⣆ ⣿⣿⠀⣰⣿⡟⢸⣿⣿⣿⡇ ⣿⣿⣿⣷⣦⠀⠀⢠⣿⣿⣿⠀⠀⣿⣿⠁⠀⣼⣿⡇⠀⢀⣴⣾⣿⡷
⠶⣶⣶⣶⣾⣿⣿⣿⣷⣶⣶⣶⠶  ⣸⣿⡟ ⠀⢠⣿⣿⠃⠈⣿⣿⠀⣿⣿⢠⣿⡿⠀⣿⣿⣧⣤⠀⢸⣿⡇⣠⣿⡿⠀⢠⣿⡟⣿⣿⠀⢸⣿⡿⠀⠀⣿⣿⠃⠀⢸⣿⣧⣄
⠀⠀⠙⢻⣿⣿⣿⣿⣿⡟⠋⠁⠀⠀ ⣿⣿⡇⠀ ⢸⣿⣿⠀⣸⣿⡟⠀⣿⣿⣾⡿⠁ ⣿⣿⠛⠛⠀⣿⣿⢿⣿⣏⠀⢀⣿⣿⣁⣿⣿⠀⣾⣿⡇⠀⢸⣿⡿⠀⠀⡀⠙⣿⣿⡆
⠀⠀⢠⣿⣿⣿⠿⣿⣿⣿⡄⠀⠀⠀ ⠙⢿⣿⣿⠇⠈⠿⣿⣿⡿⠋⠀⠀⢿⣿⡿⠁⠀⢸⣿⣿⣿⡇⢸⣿⣿⠀⣿⣿⣄⣾⣿⠛⠛⣿⣿⢠⣿⣿⣿ ⣼⣿⣿⣿ ⣿⣿⡿⠋⠀
⠀⢀⣾⠟⠋⠀⠀⠀⠙⠻⣷⡀⠀⠀
 
  v0.6.17

🔍 Detected coverage file: Build/logs/clover.xml
🚀 Posting coverage data to https://coveralls.io/api/v1/jobs
---
✅ API Response: {"message":"Job #25213017521.1","url":"https://coveralls.io/jobs/181127924"}
- 💛, Coveralls
```

https://coveralls.io/jobs/181127924

heiko-hardt:
Hi @sbuerk , could you please rebase your pull request? Unfortunately there are currently some merge conflicts. Thanks a lot!

sbuerk:
Rebased onto current `main`, and replaced rather than just rebased: the
original approach did not hold up.

It is now five commits which build on each other, meant to be extracted into
separate pull requests with their own issues. Only the fourth resolves #1802:

1. write the coverage artefacts to `.Build/` instead of `build/`
2. pass the Coveralls action its inputs — they sat in an `env:` block and were
   therefore discarded, which is a real bug on `main` today
3. collect the coverage in `runTests.sh`, via a new `-m` option and a
   `coverageMerge` suite
4. run the coverage workflow through `runTests.sh`, which is what #1802 asks for
5. drop the superseded coverage Composer scripts

All 37 required checks pass. The upload works as configured now rather than by
accident: the log says `Using coverage file: .Build/logs/clover.xml` where it
previously fell back to auto-detection, and Coveralls recorded build 81791754 at
100.0%, 126 of 126 lines.

To answer my own question from earlier in this thread: the upload from a fork
was never the problem. Coveralls does not send a comment or a status back to a
pull request from a fork, and it posts no commit status anywhere in this
repository, `main` included. A Coveralls status should therefore not be made a
required check — it would never be delivered for a contribution from a fork.

One thing deliberately not done: fanning the coverage out across the whole test
matrix. A `.cov` file carries no version information, and `phpcov` merges
whatever it is given without checking it, so reports from differently resolved
dependency sets would merge into a plausible but wrong number. The coverage is
collected for one TYPO3 and one PHP version, and `coverageMerge` refuses to run
when the collected reports do not all belong to the same one. The description
has the reasoning.

sbuerk:
A correction to my previous comment, and what happens next.

I wrote that Coveralls does not send a comment or a status back to a pull
request from a fork. That is not accurate, and the counter-evidence is in this
repository: of the ten pull requests it has ever had from a fork, two did get a
Coveralls comment (#2094 and #1936). The other eight did not, while fifteen of
fifteen recent pull requests from branches in the repository did.

What rules out the easy explanations: #2094 and #2095 were opened by the same
author on the same day, with a byte-identical workflow, and only one of them got
a comment. So it is not the fork as such, not the author, not the date and not
the configuration. The honest description is that the comment arrives reliably
for pull requests from branches in the repository and only intermittently for
pull requests from a fork.

What does hold, and is confirmed: the upload itself works from a fork. This
branch produced Coveralls builds repeatedly, the most recent at 100.0%, 126 of
126 lines. And no Coveralls commit status is delivered anywhere in this
repository, `main` included, so a Coveralls status cannot be made a required
check.

This pull request has been open since February. To find out whether the missing
comment is about the fork or about a pull request that has been open this long,
I am closing it in favour of #2336, which is the same branch with the same
commits, rebased onto current `main` and with all required checks green.

Everything discussed here is in the new pull request; this one stays readable
for the history.

PR #1989, opened 2026-03-03
[TASK] Collect all composer fix commands in `runTest.sh` fix command

### LINKED ISSUE #1855: Add `fix` command to `runTests.sh`
## Goal

As a developer, I would like to have a single command in `runTests.sh` to run all fixers (similar to `composer fix`, but without using the composer scripts).

## Acceptance Criteria

- [x] We have a command `runTests.sh -s fix` that runs all fixers that `composer fix` runs.
- [x] This is documented in the `runTests.sh` help.
- [x] This command also is checked in the `check-fixers` CI job.
- [ ] This command does not run `composer fix`.

kanow:
Resolves: #1855

### DIFF
--- a/Build/Scripts/runTests.sh
+++ b/Build/Scripts/runTests.sh
@@ -691,12 +691,11 @@ case ${TEST_SUITE} in
         SUITE_EXIT_CODE=$?
         ;;
     fix)
-        COMMAND="composer fix"
-        ${CONTAINER_BIN} run ${CONTAINER_COMMON_PARAMS} --name composer-command-${SUFFIX} -e COMPOSER_CACHE_DIR=.cache/composer -e COMPOSER_ROOT_VERSION=${COMPOSER_ROOT_VERSION} ${IMAGE_PHP} /bin/sh -c "${COMMAND}"
-        COMMAND="echo ${HELP_TEXT_NPM_CI}; npm ci --silent || { echo ${HELP_TEXT_NPM_FAILURE}; exit 1; } && npm run fix:lint:js"
-        ${CONTAINER_BIN} run ${CONTAINER_COMMON_PARAMS} --name npm-command-${SUFFIX} ${IMAGE_NODEJS} /bin/sh -c "${COMMAND}"
-        COMMAND="echo ${HELP_TEXT_NPM_CI}; npm ci --silent || { echo ${HELP_TEXT_NPM_FAILURE}; exit 1; } && npm run fix:lint:css"
-        ${CONTAINER_BIN} run ${CONTAINER_COMMON_PARAMS} --name npm-command-${SUFFIX} ${IMAGE_NODEJS} /bin/sh -c "${COMMAND}"
+        composerNormalize
+        rector
+        cgl
+        lintJs
+        lintCss
         SUITE_EXIT_CODE=$?
         ;;
     functional)


kanow:
1. Mmmh.. too much confusion for me ;-) with related #1855  Old issue with pending decision. Can we speed up that decision making?

kanow:
After talk with @oliverklee :+1: 

- asking @AnReiter which commands are still not refactored to bash scripts
  - fix command should do both, check things (with dry-run option) and fix things (is default, without dry-run option)
- the replace composer command in docs without `composer` (don't forget index)
  - there should be a list in #1855 which commands should be replaced

Build/Scripts/runTests.sh
```diff
@@ -682,12 +683,11 @@ case ${TEST_SUITE} in
...
     fix)
-        COMMAND="composer fix"
-        ${CONTAINER_BIN} run ${CONTAINER_COMMON_PARAMS} --name composer-command-${SUFFIX} -e COMPOSER_CACHE_DIR=.cache/composer -e COMPOSER_ROOT_VERSION=${COMPOSER_ROOT_VERSION} ${IMAGE_PHP} /bin/sh -c "${COMMAND}"
-        COMMAND="echo ${HELP_TEXT_NPM_CI}; npm ci --silent || { echo ${HELP_TEXT_NPM_FAILURE}; exit 1; } && npm run fix:lint:js"
-        ${CONTAINER_BIN} run ${CONTAINER_COMMON_PARAMS} --name npm-command-${SUFFIX} ${IMAGE_NODEJS} /bin/sh -c "${COMMAND}"
-        COMMAND="echo ${HELP_TEXT_NPM_CI}; npm ci --silent || { echo ${HELP_TEXT_NPM_FAILURE}; exit 1; } && npm run fix:lint:css"
-        ${CONTAINER_BIN} run ${CONTAINER_COMMON_PARAMS} --name npm-command-${SUFFIX} ${IMAGE_NODEJS} /bin/sh -c "${COMMAND}"
+        composerNormalize
+        rector
+        phpCsFixer
```
oliverklee: This should use the new name (as we'll merge `phpCsFixer` and `cgl`):


```suggestion
        cgl
```

oliverklee (changes requested):

Build/Scripts/runTests.sh
kanow: @oliverklee This (`cgl`) runs without a bash command?
I didn't see a bash command for that.

kanow: cgl is now a bash function and will be used in the `fix` command.

Build/Scripts/runTests.sh
```diff
@@ -354,6 +354,14 @@ EOF
 }
 
 # Functions for the individual checkers/fixers
+cgl() {
```
oliverklee: Let's move this to a separate (pre-)PR, please. Also, we should then avoid calling the Composer scripts and copy what the Core does:

https://github.com/TYPO3/typo3/blob/main/Build/Scripts/runTests.sh#L1045

oliverklee (changes requested):

Build/Scripts/runTests.sh
oliverklee: #2259

kanow: Other PR is merged

oliverklee (approved):

Issue #1867, opened 2025-11-10
Add a BE module


oliverklee:
## Goal

As a tea lover, I'd like to conveniently see all my teas in a BE module.

As a developer, I'd like to see a working example of an Extbase BE module and functional tests that I can copy from so that I can easily have well-tested BE modules in my own extension.

## Acceptance Criteria
- [x] We have a BE module "Tea". (https://github.com/TYPO3BestPractices/tea/pull/2022)
- [x] The BE module has the Tea icon. (https://github.com/TYPO3BestPractices/tea/pull/2022)
- [x] The BE module lists the tea records from the whole installation. (https://github.com/TYPO3BestPractices/tea/pull/2022)
- [x] The BE module shows the UID and title of each tea. (https://github.com/TYPO3BestPractices/tea/pull/2022)
- [x] If there are no teas, we have a corresponding flash message. (https://github.com/TYPO3BestPractices/tea/pull/2029)
- [x] The list is sorted by UID, highest first. (https://github.com/TYPO3BestPractices/tea/pull/2032)
- [x] The title and UID are linked to the TCEforms edit action. (https://github.com/TYPO3BestPractices/tea/pull/2036)
- [x] Going back from editing a tea shows the tea module again. (https://github.com/TYPO3BestPractices/tea/pull/2036)
- [x] We have a follow-up ticket for adding a full text search. (https://github.com/TYPO3BestPractices/tea/issues/2039)
- [x] We have a follow-up ticket for adding a delete button. (https://github.com/TYPO3BestPractices/tea/issues/2038)

PR #2022, opened 2026-03-26
[FEATURE] Add backend module to demonstrate testing of them

DanielSiepmann:
We add a very basic version of a backend module written using extbase. The goal is to provide an working example on how to test backend modules written with extbase.

Relates: #1867


Tests/Functional/Controller/BackendModuleControllerTest.php
```diff
@@ -0,0 +1,151 @@
...
+        $html = $response->getBody()->__toString();
+
+        self::assertGreaterThan(
+            strpos($html, 'Tea 1'),
+            strpos($html, 'Tea 2'),
+            'Tea 1 is not sorted after Tea 2'
+        );
+    }
+
+    private function executeRequest(
```
oliverklee: We probably can annotate the parameters as `non-empty-string` using PHPDoc.

PR #2043, opened 2026-03-27
[TASK] Add PHP CS Fixer to `runTests.sh`

AnReiter:


Build/Scripts/runTests.sh
```diff
@@ -205,6 +205,7 @@ Options:
             - lintTypoScript: TypoScript linting
             - lintYaml: YAML linting
             - npm: "npm" with all remaining arguments dispatched.
+            - php-cs-fixer fixes code to follow the standards.
```
oliverklee: I've noticed that my earlier suggestion for the name (kebab case) does not match the current naming conventions for the `runTests.sh` commands (lower camelCase). So this should be `phpCsFixer`. I'm sorry for the double work. 😔

PR #2054, opened 2026-03-28
[CLEANUP] Autoformat the code

oliverklee:


DanielSiepmann (commented):
Is there any way to automate this via CI?
I fear we would otherwise always result in mixed formats within the extension.

PR #2066, opened 2026-03-28
[DOCS] Add new pages for the setup in three different use cases to documentation

Schira-N:
This is about creating three different setup pages. 
In follow-up PRs, the content of the pages will be reworked.

### DIFF
--- a/Documentation/ForCodeExplorers.rst
+++ b/Documentation/ForCodeExplorers.rst
@@ -10,3 +10,7 @@ Your use case: you just want to have a look into the code and find examples of b
 
 The easiest way is to head to github and inspect the code online:
 
+.. code-block:: php
+
+  https://github.com/TYPO3BestPractices/tea
+
--- a/Documentation/ForCodeExplorers.rst
+++ b/Documentation/ForCodeExplorers.rst
@@ -1,0 +1,12 @@
+.. include:: /Includes.rst.txt
+
+.. _forCodeExplorers:
+
+==================
+For Code-Explorers
+==================
+
+Your use case: you just want to have a look into the code and find examples of best-practices.
+
+The easiest way is to head to github and inspect the code online:
+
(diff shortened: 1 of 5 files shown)


oliverklee (changes requested):
Please let's add them one by one. That way, we can keep the PRs smaller and merge the changes faster.

Schira-N:
Sorry, I don't see the reason, why commits of documentation should be split into tiny chunks. 
For me, even the existing commit is needlessly small. 
I think, textual changes have to be handled differently than code changes. 

I would rather commit the whole changed documentation in a single commit. 

The reason: Only then the full picture is visible and only then sensible improvements can be applied. (I am absolutely open to further improving the documentation and changing things to the better!)

Working with texts goes like this: The autor creates a text. The proofreader reads through the whole text and creates a bunch of questions, suggestings and change requests. The autor, than creates a new version of the text, based on all the comments. This circle will be continued until both are satisfied.

Of course: The proofreader will need some time to read through the whole stuff. In our case, he needs to create the generated documentation first and commenting on that is technically not so easy, I admit. Maybe, using screenshots with added notes would be a way to go... Doing a review this way will require surely 30min or more.

oliverklee:
After having a quick glance at these changes, I think we still need to make some changes both to the naming of the parts as well as the contents of the parts. And review time/effort roughly grows quadratically with the size of the change.

So I'd prefer to split this PR up so we can get the separate parts merged faster.

(If you're not willing to split this up, we'll need to bite the bullet, put more time into the reviews, and merge the changes later.)

oliverklee:
We could also add almost-empty parts as placeholders in a separate PRs first. That way, we can work out the naming first without also having to think about the specific contents.

heiko-hardt:
Hi @Schira-N , could you please rebase your pull request? Unfortunately there are currently some merge conflicts. Thanks a lot!

PR #2092, opened 2026-04-21
[CLEANUP] Move `check:php:lint` before `check:php:cs-fixer` in Composer

AnReiter:
When checking the code, the PHP linting always needs to come before other PHP-related checkers (PHP-CS-Fixer, Rector, PHPStan, PHP_CodeSniffer) as there's no point in checking anything else if the PHP is not valid.


oliverklee (approved):
Thanks!

PR #2094, opened 2026-04-27
[BUGFIX] Update Node.js and NPM engine version requirements

### LINKED ISSUE #2071: Fix `package.json` to unblock Dependabot updates
https://github.com/TYPO3BestPractices/tea/network/updates/1338329726

Dependabot uses Node.js v24.14.1 and NPM 11.11.0. Due to the [engine-strict setting](https://docs.npmjs.com/cli/v8/using-npm/config#engine-strict), the update will not succeed.

We might need to update some of the used packages.

drc-ravisachaniya:
Resolves: #2071


package.json
```diff
@@ -30,4 +30,4 @@
 		"stylelint-config-standard-scss": "^17.0.0",
 		"stylelint-no-browser-hacks": "^2.0.0"
 	}
-}
+}
```
oliverklee:
```suggestion
}

```

oliverklee (approved):
Thank you!

I'll start Dependabot after merging to see if the issue then is fixed.

oliverklee:
Ah, no, unfortunately, this didn't solve the problem: https://github.com/TYPO3BestPractices/tea/network/updates/1338329726

drc-ravisachaniya:
I think `.nvmrc` file also need to update the node version

PR #2096, opened 2026-04-27
[TASK] Undo NPM/Node.js version requirements

oliverklee:
The change did not help with Dependabot updates and should be reverted.

This reverts commit 21b5ae6345efd7bc363c626701f781213cdbc254.

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -20,8 +20,8 @@
 				"stylelint-no-browser-hacks": "^2.0.0"
 			},
 			"engines": {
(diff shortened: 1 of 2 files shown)


AnReiter:
could you be more specific? :)
and rebase!

oliverklee:
> could you be more specific? :)

The reverted commit was supposed to fix #2071, i.e., to allow Dependabot to update the NPM packages without errors. However, the commit did not help. So we should revert it and implement a working fix.

AnReiter (approved):

PR #2107, opened 2026-05-01
[TASK] Add basic unit and functional tests

sbuerk:
This change adds following tests:

* `Tests/Functional/ExtensionLoadedTest.php` to verify that
   extension is detected as loaded with expected extension
   key and composer package name. On top it tests against
   expected TYPO3 core versions using group excludes, which
   helps to simpler detect if tests for one core version
   are executed with the other version installed.

*  `Tests/Unit/VersionCompatTest.php`  tests against expected
   TYPO3 core versions using group excludes, which  helps to
   simpler detect if tests for one core version are executed
   with the other version installed.

In general these two tests can act as first tests for any
extension to have unit and functional test infrastructure
added. Having at least the functional tests helps detecting
DI container compile issues or automatic TCA migrations and
related deprecations early - even if no other thats are added.

For example, if `12.4` has been used to install the dependencies
and the unit/functional tests are executed giving the `-t 13.4`
option these tests will fail and indicates this - other failures
can than be early ruled out and avoids wasting debug time

`ci.yaml` is modified to call unit and functional test with the
same TYPO3 version used to setup the system.

> [!NOTE]
> Existing similar tests are removed in favour of the new ones,
> and having them on the lowest level makes them easy findable.


oliverklee (changes requested):
We already have some similar tests:

- https://github.com/TYPO3BestPractices/tea/blob/main/Tests/Functional/Environment/ExtensionTest.php
- https://github.com/TYPO3BestPractices/tea/blob/main/Tests/Unit/Environment/ExtensionTest.php

Let's drop or merge them.

oliverklee:
I've created a pre-PR for moving and renaming the testcases: https://github.com/TYPO3BestPractices/tea/pull/2108

Tests/Unit/VersionCompatTest.php
```diff
@@ -0,0 +1,42 @@
...
+use TYPO3\CMS\Core\Information\Typo3Version;
+use TYPO3\TestingFramework\Core\Unit\UnitTestCase;
+
+/**
+ * Basic unit tests.
+ *
+ * This can and should be used in every extension instead of dummy assertTrue(true) filler.
+ */
+#[CoversNothing]
+final class VersionCompatTest extends UnitTestCase
```
oliverklee, line 19: Let's avoid the abbreviation:

```suggestion
final class VersionCompatibiltyTest extends UnitTestCase
```

oliverklee (changes requested):
(not a full review yet, and would like the pre-PR to get merged first)

heiko-hardt:
Hi @sbuerk , could you please rebase your pull request? Unfortunately there are currently some merge conflicts. Thanks a lot!

sbuerk:
Superseded by #2327 and closed in favour of it.

This pull request was opened from a fork and had grown merge conflicts
with `main`. Instead of rebasing it there, the change has been taken over
into a branch of this repository, so it can be maintained here directly.

The review feedback of this pull request is covered:

* The near-duplicate test cases are gone. They were moved and renamed with
  #2108 in the meantime, so the class name no longer carries the
  abbreviation, and the new tests extend those classes instead of adding
  further ones.

The change itself needed a rework for the current state of the repository:

* The supported TYPO3 versions are 13.4 and 14.3 now, so the pinned tests
  and their groups apply to those instead of 12.4 and 13.4.
* The comma separated value for `--exclude-group` is rejected by
  PHPUnit 11.5 with a test runner warning. Together with
  `failOnAllIssues="true"` in the PHPUnit configuration this fails the
  whole functional suite regardless of the test results, so the two
  groups are passed as two separate options.
* With that, the file wide `SC2054` exception for shellcheck is not
  needed any more.
* `weekly.yml` runs the same unit and functional tests and needed `-t`
  as well, otherwise its TYPO3 14.3 jobs would have run as 13.4 and
  failed on the pinned test.
* The Composer scripts which call PHPUnit directly have no TYPO3 version
  to select, so they exclude the pinned tests of both versions rather
  than the ones of a single hardcoded version.
* The `-t` option is documented for the test suites now, in the help text
  of `runTests.sh` and in the documentation, and a failing pinned test
  states which version the run was made for.

Verified locally for TYPO3 13.4 and 14.3 with PHP 8.2: static checks,
PHPStan, unit and functional tests pass for both, and a run made for the
other version fails with the pinned test in all four combinations.

PR #2109, opened 2026-05-01
[TASK] Move PHPCov from Phive to Composer

oliverklee:
Now that we require PHP >= 8.1, we can use PHPCov as a Composer dependency again.

This allows us to stop using PHIVE in addition to Composer as a dependency manager.

```
./Build/Scripts/runTests.sh -s composer req --dev phpunit/phpcov:9.0.2
```


AnReiter:
looks good!

AnReiter (approved):

PR #2113, opened 2026-05-04
[TASK] Move `phpCsFixer` to a bash function

AnReiter:


oliverklee (approved):
This works fine:

- [x] detect and show issues with `./Build/Scripts/runTests.sh -s phpCsFixer -- --dry-run`
- [x] exit code != 0 if there are issues
- [x] exit code = 0 if there are no issues

This doesn't work on my machine (but also was broken before):

- [ ] actually fix the code using `./Build/Scripts/runTests.sh -s phpCsFixer`

I'll add a ticket for that.

Issue #1711, opened 2025-06-23
Check if we should adapt our RST rendering check to what the Core does


oliverklee:
https://review.typo3.org/c/Packages/TYPO3.CMS/+/89790

kanow:
@oliverklee not sure what you mean. Do we need a check for the rendered documentation? Or we just should adapt the processing command and add some more `echo` to make it more understandable?

oliverklee:
@kanow I've just re-checked both Core changes:
- The first change doesn't look relevant for us after all.
- Let's have a look at the second change (or even better: the [current state in the Core `runTests.sh`](https://github.com/TYPO3/typo3/blob/main/Build/Scripts/runTests.sh#L313)), compare it against what we currently have, and see what we'd like to adopt from there. (We might also want to adopt the command name to better be in sync with the Core `runTests.sh`.)

kanow:
Resolved and mergd

PR #2115, opened 2026-05-04
[TASK] Adapt command to render documentation

kanow:
Improve naming of the command and some variables,
remove unnecessary option.

Related: #1711


oliverklee (changes requested):
We'll also need to adapt `Documentation/TechnicalBackground/Documentation.rst` to the changed command.

oliverklee (commented):
Works fine, BTW. Thanks! Only need to update the docs then.

kanow:
docs now updated

oliverklee (approved):

PR #2136, opened 2026-05-21
[BUGFIX] Actually fix the code with `runTests.sh -s phpCsFixer`

### LINKED ISSUE #2120: `runTests.sh -s phpCsFixer` doesn't actually fix the code
## Summary

`runTests.sh -s phpCsFixer` doesn't fix fixable code style warnings.

1. Move the `{` in a PHP class to the end of the previous line.
2. Run `runTests.sh -s phpCsFixer`.
3. Check whether the code is fixed.

## Actual behavior

The code is unchanged.

## Expected behaviour

The code is fixed.

oliverklee:
Also align the dry-run handling with what the Core `runTests.sh` now does.

Fixes #2120


kanow (approved):
Tested that locally, it works fine.

PR #2148, opened 2026-06-01
[FEATURE] Use PHP script for XLIFF check in `runTests.sh`

kanow:
Same as the core does the xliff check.


kanow:
I checked the functionality of the old `lintXliff` check. IMO the new che does all the things that the old lint command did. Therefore I removed the old `lintXliff` command.
I'm not sure about the naming of the composer script. For the moment i didn't change the script naming for composer. The name in  github workflow is therefore also not changed.

Build/Scripts/checkIntegrityXliff.php
```diff
@@ -0,0 +1,318 @@
+#!/usr/bin/env php
```
oliverklee: We're calling this with `php checkIntegrityXliff.php`. So I don't think we need this line.

Build/Scripts/checkIntegrityXliff.php
```diff
@@ -0,0 +1,318 @@
...
+<?php
+
+declare(strict_types=1);
+
+/*
+ * This file is part of the TYPO3 CMS project.
```
oliverklee: This the Core header. It's not applicable to extensions, and we should remove it.

Build/Scripts/checkIntegrityXliff.php
```diff
@@ -0,0 +1,318 @@
...
+        $isVerbose = in_array('-v', $argv, true) || in_array('--verbose', $argv, true);
+
+        $filesToProcess = $this->findXliff();
+        $output = new ConsoleOutput();
+        $output->setFormatter(new OutputFormatter(true));
+
+        $testResults = [];
+        $errors = [];
+
+        /** @var \SplFileInfo $labelFile */
```
oliverklee: Please let's convert this to an `assert()` to enforce the type.

composer.json
```diff
@@ -45,7 +45,8 @@
 		"typo3/cms-extbase": "^12.4.41 || ^13.4",
 		"typo3/cms-fluid": "^12.4.41 || ^13.4",
 		"typo3/cms-frontend": "^12.4.41 || ^13.4",
-		"typo3/cms-install": "^12.4.41 || ^13.4"
+		"typo3/cms-install": "^12.4.41 || ^13.4",
+		"typo3/minimal": "^13.4"
```
oliverklee: We need to remove this line again.

composer.json
```diff
@@ -66,7 +67,9 @@
 		"spaze/phpstan-disallowed-calls": "4.12.0",
 		"ssch/typo3-rector": "2.15.2 || 3.14.1",
 		"ssch/typo3-rector-testing-framework": "2.0.1 || 3.0.0",
+		"symfony/config": "^6.4 || ^7.4 || ^8.1",
 		"symfony/console": "^6.4 || ^7.4",
+		"symfony/finder": "^6.4 || ^7.4 || ^8.1",
```
oliverklee: As long as we don't support TYPO3 14LTS, we don't need Symfony 8.1 for these added dependencies.

oliverklee (changes requested):

Build/Scripts/checkIntegrityXliff.php
kanow: @oliverklee Sorry, I have no idea what I should do here ;-)

oliverklee: > Sorry, I have no idea what I should do here ;-)

Replace the PHPDoc type annotation with this as first line in the loop:

```php
assert($labelFile instanceof \SplFileInfo);
```

kanow:
@oliverklee I have no idea why the composerUnused check fails with the message that `typo3/cms-fluid` is an unused package!? That was already there before my changes here. Why it fails now? Any idea?

oliverklee:
@kanow You'll need to re-add the `typo3/cms-install` to the dependencies first. (It's needed for the upgrade wizard.)

kanow:
Too much clean up ;-)
@oliverklee done, checks green

oliverklee (approved):

PR #2153, opened 2026-06-01
[CLEANUP] Drop redundant return statements from `runTests.sh`

oliverklee:


kanow (approved):

AnReiter:
no i think important for Exitcodes for the tests. or not?

oliverklee:
> no i think important for Exitcodes for the tests. or not?

I've manually tested that the exit code still gets returned even if we drop the `return`.

PR #2166, opened 2026-06-01
[CLEANUP] Autoformat the XLIFF files

oliverklee:


kanow:
@oliverklee It looks to me there are no rules for moving code in the next line. A little bit confusing. Sometimes the original is starting the new line, sometimes the date. Is that normal? Should that be in this way?

oliverklee:
> It looks to me there are no rules for moving code in the next line. A little bit confusing. Sometimes the original is starting the new line, sometimes the date. Is that normal? Should that be in this way?

I think PhpStorm is doing this:

"Fit as many things as possible on the line, and then break before the allowed maximum length (as configured in PhpStorm) is reached."

kanow:
If this depends on configuration of PhpStorm (or another IDE) then, maybe if another user is formatting the code, we have the next change? I'm not sure that we should change it here. What is the point of this change? Is this kind of "Best Practice"? Probably, yes. And if you like it, then I don't have a problem with that. Just want to thinking and talk about it.

oliverklee:
>  I'm not sure that we should change it here. What is the point of this change?

When I a developer edits a file, they should be able to autoformat it with the default PhpStorm settings to clean up their changes without introducing unrelated formatting changes.

kanow:
Ok.

kanow (approved):

PR #2199, opened 2026-07-06
[TASK] Update most GitHub Action runners to Ubuntu 26.04

oliverklee:
For the code coverage workflow (which does not use `runTests.sh` yet), we stay at 24.04 as the Ubuntu 26.04 GitHub Actions runner does not provide MySQL out of the box anymore.

We'll update to 26.04 after we've switch the code coverage CI workflow to use `runTests.sh`.

### DIFF
--- a/.github/workflows/auto-approve.yml
+++ b/.github/workflows/auto-approve.yml
@@ -11,7 +11,7 @@ jobs:
   auto-approve:
     name: Dependabot auto-approve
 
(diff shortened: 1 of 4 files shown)


oliverklee:
Note to myself: https://github.com/marketplace/actions/actions-setup-mysql

Starkmann (approved):

PR #2211, opened 2026-07-06
[TASK] Drop the `ListTypeToCTypeUpdate` upgrade wizard

oliverklee:
All 4.x versions of this extension have had this upgrade wizard. So we can expect all users to have run the upgrade, and we can remove the upgrade wizard.

We can also remove the `typo3/cms-install` dependency, which was only relevant for the upgrade wizard.

```
composer rem typo3/cms-install
```


kanow (approved):
Looks good to me and make sense to remove that upgrade wizard.

PR #2241, opened 2026-08-03
[DOCS] Change composer command to use `runTests.sh` to run all fixers

kanow:
Related: #1855

If I see that right https://github.com/TYPO3BestPractices/tea/issues/1855#issuecomment-4370244102 the `fix` command in the documentation should be changed.
The fix command is now complete.

### DIFF
--- a/Documentation/TechnicalBackground/Running.rst
+++ b/Documentation/TechnicalBackground/Running.rst
@@ -118,10 +118,10 @@ Lints the TypoScript files.
 
 Lints the YAML files.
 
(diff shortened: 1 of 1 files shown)


Documentation/TechnicalBackground/Running.rst
```diff
@@ -121,7 +121,7 @@ Lints the YAML files.
 .. index:: Commands; composer fix
```
oliverklee: We should update the index entries as well.

oliverklee (approved):

PR #2242, opened 2026-08-03
[TASK] Move `checkComposerNormalize` to a bash function

AnReiter:


Build/Scripts/runTests.sh
```diff
@@ -356,6 +356,11 @@ EOF
 
 # Functions for the individual checkers/fixers
 
+checkComposerNormalize() {
```
oliverklee: This needs to be `composerNormalize` instead and allow an optional dry-run.

oliverklee (changes requested):

Build/Scripts/runTests.sh
```diff
@@ -623,8 +631,7 @@ case ${TEST_SUITE} in
         SUITE_EXIT_CODE=$?
         ;;
     checkComposerNormalize)
-        COMMAND="composer normalize --no-check-lock --dry-run"
-        ${CONTAINER_BIN} run ${CONTAINER_COMMON_PARAMS} --name composer-normalize-${SUFFIX} -e COMPOSER_CACHE_DIR=.cache/composer -e COMPOSER_ROOT_VERSION=${COMPOSER_ROOT_VERSION} ${IMAGE_PHP} /bin/sh -c "${COMMAND}"
+        composerNormalize
```
oliverklee, line 635: As long as this `runTests.sh` command is called `checkComposerNormalize`, we need to make sure the bash function is run in dry-run mode:


```suggestion
        CGLCHECK_DRY_RUN="-n"
        composerNormalize
```

oliverklee (changes requested):

PR #2243, opened 2026-08-03
[TASK] Use PHP strict mode for all configuration files

oliverklee:


heiko-hardt (approved):
Looks good, no objections from my side.

PR #2246, opened 2026-08-03
[TASK] Add optional dry-run to composerNormalize

AnReiter:


.github/workflows/ci.yml
```diff
@@ -137,7 +137,7 @@ jobs:
       fail-fast: false
       matrix:
         command:
-          - "checkComposerNormalize"
+          - "composerNormalize"
```
oliverklee, line 140: We'll also need to update the weekly CI workflow in `weekly.yml` so both workflows stay in sync. (Unfortunately, I haven't been able to find an easy way for both workflows to share the code.)

Build/Scripts/runTests.sh
```diff
@@ -622,8 +622,11 @@ case ${TEST_SUITE} in
         ${CONTAINER_BIN} run ${CONTAINER_COMMON_PARAMS} --name composer-command-${SUFFIX} -e COMPOSER_CACHE_DIR=.cache/composer -e COMPOSER_ROOT_VERSION=${COMPOSER_ROOT_VERSION} ${IMAGE_PHP} /bin/sh -c "${COMMAND}"
         SUITE_EXIT_CODE=$?
         ;;
-    checkComposerNormalize)
-        COMMAND="composer normalize --no-check-lock --dry-run"
+    composerNormalize)
```
oliverklee: This command then needs to go further down so the commands stay sorted in alphabetical order.

We'll also need to update the help text (`runTests.sh -s help`).

Also, we then can and should drop the `fixComposerNormalize` command, update all calls, and remove it from the help.

oliverklee (changes requested):

kanow:
Maybe @oliverklee should do the review because this contradicts itself with the PR I merged before from @heiko-hardt ?

https://github.com/TYPO3BestPractices/tea/pull/2248

oliverklee (approved):

PR #2287, opened 2026-08-31
[TASK] Do not automatically add `#[DoesNotPerformAssertion]`

oliverklee:
The TYPO3 testing framework always has some assertions in its `tearDown()`, which Rector does not detect. So Rector should not automatically add the `#[DoesNotPerformAssertion]` attribute to tests that do not contain any assertions.


AnReiter:
rebase :)

AnReiter (approved):

PR #2288, opened 2026-08-31
[TASK] Make Rector a direct development dependency

### LINKED ISSUE #2285: Make Rector a direct development dependency
Recently, the CI build has been failing due to issues with newer Rector versions. As long as Rector is only a transitive (development) dependency (of typo3-rector), this will keep happening. So let's make Rector a direct development dependency, and pin it on the last version that works.

oliverklee:
So far, `rector/rector` has been a transitive dependency
of the `ssch/typo3-rector` package. This allowed
updates of `rector/rector` to happen without us touching
the `composer.json`, and causing CI breakage.

In order to keep the specific version of `rector/rector`
under strict control, it now is a direct development
dependency with a pinned version.

```
composer req --dev rector/rector:2.6.3
```

Fixes #2285

### DIFF
--- a/composer.json
+++ b/composer.json
@@ -60,6 +60,7 @@
 		"phpstan/phpstan-strict-rules": "2.0.12",
 		"phpunit/phpcov": "10.0.1",
 		"phpunit/phpunit": "11.5.56",
+		"rector/rector": "2.6.3",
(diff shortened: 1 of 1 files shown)


heiko-hardt (approved):
Looks good, no objections from my side.

PR #2303, opened 2026-09-07
[FEATURE] Add the image to the show action

### LINKED ISSUE #2298: Add the image to the show action
## Acceptance criteria

- [ ] between the title and the description
- [ ] cover it with functional tests

AnReiter:
The single view rendered the title and the description of a tea, but
not its image. The image is now rendered between the two, wrapped in a
`figure` element and limited to a width of 600 pixels so that a large
image cannot break the layout. The alternative text of the file
reference is rendered as the `alt` attribute for accessibility.

The functional tests cover the rendered image, its alternative text and
its scaled width, and that no `figure` element is rendered for a tea
without an image.

The testing framework configures GraphicsMagick as the image processor
of every functional test instance. The container images used by
`runTests.sh` ship it, but the runner image of the code coverage
workflow does not. Image processing there silently falls back to the
original file, which is then rendered without any dimensions, so that
workflow now installs GraphicsMagick.

Two repository tests are marked with `IgnoreDeprecations` on top of
that. The `Validate` attributes of the model trigger an Extbase
deprecation on TYPO3 14.3, which fails the functional test suite there.
A `@todo` records that this is unrelated to this change and needs an
investigation of its own.

Resolves: #2298


Resources/Private/Templates/Tea/Show.html
```diff
@@ -1,12 +1,18 @@
 <!DOCTYPE html>
 <html xmlns:f="http://typo3.org/ns/TYPO3/CMS/Fluid/ViewHelpers" data-namespace-typo3-fluid="true">
-    <f:layout name="Default"/>
+<f:layout name="Default"/>
```
oliverklee: Let's get #2307 in first and then reformat to avoid the unrelated formatting changes.

Resources/Private/Templates/Tea/Show.html
```diff
@@ -1,12 +1,18 @@
...
+<f:section name="main">
+    <h2>
+        {tea.title}
+    </h2>
 
-        {tea.description -> f:format.html()}
-    </f:section>
+    <f:if condition="{tea.image}">
+        <figure>
+            <f:image image="{tea.image}"/>
```
oliverklee: We need to have the `alt` attribute for accessibility. Also, we should set a sensible `maxWidth` for the image in order to not break the layout.

oliverklee (changes requested):

Tests/Functional/Controller/TeaControllerTest.php
```diff
@@ -107,6 +111,44 @@ public function showActionRendersTheGivenTeas(): void
...
+    #[Test]
+    public function showActionRendersMaxWidthFromImageOfTea(): void
+    {
+        $this->importCSVDataSet(__DIR__ . '/Fixtures/Database/TeaController/TeaWithImage.csv');
+
+        $request = (new InternalRequest())->withPageId(3)->withQueryParameters(['tx_tea_teashow[tea]' => 1]);
+
+        $html = (string)$this->executeFrontendSubRequest($request)->getBody();
+
+        self::assertStringContainsString('width="600', $html);;
```
oliverklee: It looks like the closing double quote is missing.

oliverklee (changes requested):
This also should get a changelog entry.

sbuerk:
Rebased onto current `main` and updated.

The branch was rebased rather than merged, so the merge commit that the
"Update branch" button had added is gone and the history is linear again.

Changes since the last review:

* Added the changelog entry.
* Wrote the `f:image` tag on a single line, matching the other view helpers in
  the templates and the HTML settings added in #2307.
* Fixed the failing `Calculate code coverage` job. The testing framework
  configures GraphicsMagick as the image processor of every functional test
  instance. The container images used by `runTests.sh` ship it, the runner
  image of that workflow does not, so image processing silently fell back to
  the original file and rendered ``. That
  workflow now installs GraphicsMagick. It is also why the job failed while
  the functional test jobs passed on the very same commit.
* Renamed the tea in both fixtures from `ImageOfTea` to `Earl Grey`. The title
  was the same string as the image file name, so the assertion on
  `ImageOfTea` could have been satisfied by the rendered title alone instead
  of by the image.
* Added a test that no `figure` element is rendered for a tea without an
  image, which covers the other branch of the `f:if`.
* Marked `findAllSortsByTitleInAscendingOrder` and
  `findByUidForExistingRecordReturnsModel` with `IgnoreDeprecations`. The
  `Validate` attributes of the model trigger an Extbase deprecation on TYPO3
  14.3 that failed `Functional tests (14.3, 8.5, sqlite, Min)`. A `@todo`
  records that this is unrelated to this change and needs an investigation of
  its own.

Run locally for both cores before pushing: PHPStan, unit and functional tests
on 13.4 (SQLite, MySQL, PostgreSQL) and on 14.3 (SQLite), plus the lowest
dependency set of 14.3 that had failed.

CI is green: 43 checks pass, the remaining two are the skipped Dependabot jobs.

oliverklee:
@sbuerk Thanks!

CHANGELOG.md
```diff
@@ -7,6 +7,7 @@ This project adheres to [Semantic Versioning](https://semver.org/).
 
 ### Added
 
+- Add the image to the show action (#2298)
```
oliverklee:
```suggestion
- Add the image to the show action (#2303)
```

oliverklee (approved):

PR #2306, opened 2026-09-07
[BUGFIX] Avoid changing autoload files when checking PSR conformity

### LINKED ISSUE #1948: Unit and functional tests cannot determine the entry script after `composer check:composer:psr-verify`
## Summary

Unit and functional tests fail with an error message if `composer check:composer:psr-verify` has been run and `composer update` has not been run after that.

```bash
./Build/Scripts/runTests.sh -u
./Build/Scripts/runTests.sh -s clean
./Build/Scripts/runTests.sh -s composer up -W
./Build/Scripts/runTests.sh -s composer check:composer:psr-verify
./Build/Scripts/runTests.sh -s unit
```

## Actual behavior

```
Unable to determine path to entry script. Please check your path or set an environment variable 'TYPO3_PATH_ROOT' to your root path.
ERRO[0000] Removing container 1e2e421b0a54497e865f21b1dea4358ddb9a5ae07648135f33265100c1204e6b: cleaning up container 1e2e421b0a54497e865f21b1dea4358ddb9a5ae07648135f33265100c1204e6b: removing container 1e2e421b0a54497e865f21b1dea4358ddb9a5ae07648135f33265100c1204e6b network: 1 error occurred:
        * rootless netns: kill network process: permission denied
 

###########################################################################
Result of unit
Container runtime: podman
Container suffix: 16028
PHP: 8.2
TYPO3: 12.4
FAILURE
###########################################################################
```

## Expected behaviour

```
PHPUnit 10.5.63 by Sebastian Bergmann and contributors.

Runtime:       PHP 8.2.30
Configuration: /home/klee/src/typo3/ext/tea/Build/phpunit/UnitTests.xml

.........                                                           9 / 9 (100%)

Time: 00:00.004, Memory: 12.00 MB

OK (9 tests, 36 assertions)
ERRO[0000] Removing container a9996ccfe32bb68393fab3c16a431ba13077ade3f5785996c3372ce4eb19941f: cleaning up container a9996ccfe32bb68393fab3c16a431ba13077ade3f5785996c3372ce4eb19941f: removing container a9996ccfe32bb68393fab3c16a431ba13077ade3f5785996c3372ce4eb19941f network: 1 error occurred:
        * rootless netns: kill network process: permission denied
 

###########################################################################
Result of unit
Container runtime: podman
Container suffix: 2705
PHP: 8.2
TYPO3: 12.4
SUCCESS
###########################################################################

```

## Additional information

(additional information that might help understand and fix this bug)

cumuru:
The verifyPsr script (both in runTests.sh and as a composer script) modified the autoload files in a way that subsequent script calls like e.g. unit failed. This violates the expectation of check scripts not modifying test results.
To fix the issue, an additional composer install call has been introduced in both places. It's quiet to avoid cluttering script output.

Fixes #1948

### DIFF
--- a/composer.json
+++ b/composer.json
@@ -122,7 +122,7 @@
 	},
 	"scripts": {
 		"check:composer:normalize": "@composer normalize --no-check-lock --dry-run",
-		"check:composer:psr-verify": "@composer dumpautoload --optimize --strict-psr --no-plugins",
+		"check:composer:psr-verify": "@composer dumpautoload --optimize --strict-psr --no-plugins --dry-run",
 		"check:composer:unused": "composer-unused --configuration=Build/composer-unused/composer-unused.php",
 		"check:coverage": [
 			"@check:coverage:unit",
--- a/Build/Scripts/runTests.sh
+++ b/Build/Scripts/runTests.sh
@@ -440,7 +440,7 @@ phpstanGenerateBaseline() {
 }
 
 psrVerify() {
-    COMMAND="composer dumpautoload --optimize --strict-psr --no-plugins"
+    COMMAND="composer dumpautoload --optimize --strict-psr --no-plugins --dry-run"
     ${CONTAINER_BIN} run ${CONTAINER_COMMON_PARAMS} --name psr-verify-${SUFFIX} ${IMAGE_PHP} ${COMMAND}
 }
 


composer.json
```diff
@@ -122,7 +122,10 @@
 	},
 	"scripts": {
 		"check:composer:normalize": "@composer normalize --no-check-lock --dry-run",
-		"check:composer:psr-verify": "@composer dumpautoload --optimize --strict-psr --no-plugins",
+		"check:composer:psr-verify": [
+			"@composer dumpautoload --optimize --strict-psr --no-plugins",
+			"@composer install --quiet"
```
oliverklee: This would install the latest packages, i.e., it would undo previous installs with the lower TYPO3 version (13 instead of 14) and installations with the minimal allowed version. We don't want that.

composer.json
```diff
@@ -122,7 +122,10 @@
 	},
 	"scripts": {
 		"check:composer:normalize": "@composer normalize --no-check-lock --dry-run",
-		"check:composer:psr-verify": "@composer dumpautoload --optimize --strict-psr --no-plugins",
```
oliverklee, line 125: However, I've discovered that the `dumpautoload` command has a `--dry-run` switch that still does the PSR4 check, but doesn't write anything, i.e., it does not have the side effect of breaking subsequent test runs.

oliverklee (changes requested):

composer.json
cumuru: Nice! I‘ve changed the patch accordingly.

oliverklee (approved):

PR #2327, opened 2026-09-16
[TASK] Pin the tests to the installed TYPO3 version

sbuerk:
The unit and the functional tests ran against whatever TYPO3 version
happened to be installed. The option `-t` of `runTests.sh` only selected the
version to install the dependencies for and the PHPStan configuration, so a
test run made for one version against the dependencies of the other failed
somewhere in the middle of the suite, where the cause is hard to recognize.

The new `Tests/Support/VersionCompatibilityTestsTrait.php` holds the tests
which verify the TYPO3 version, and both the unit and the new functional
`VersionCompatibilityTest` use it, so the two suites cannot drift apart. It is
annotated with `@phpstan-require-extends`, so that it can only be used by a
PHPUnit test case. Besides the check that the installed version is a supported
one, it holds one test pinned to each of the supported versions by a
`not-core-<version>` group. `runTests.sh` excludes the groups which do not
belong to the version given with `-t`, so exactly the test of the selected
version remains and fails right away if another version is installed, with a
message naming the version the run was made for and how to install the
matching dependencies.

The unit and the functional tests are called with `-t` in both the push and
the weekly workflow, so they run for the version their dependencies have been
installed for. The Composer scripts call PHPUnit directly and have no version
to select, so they exclude the pinned tests of both versions. The check for a
supported version has no group and therefore runs in all of these cases.

The help text of `-t` and the documentation state now that the option applies
to the test suites as well.

The functional tests additionally cover that the extension is recognized by
its Composer package name, and the test for the extension key now says so in
its name.

The change comes from pull request #2107, which was opened from a fork, and
has been reworked for the current state of the repository: the test cases it
moved have been moved with #2108 in the meantime, the supported versions are
13.4 and 14.3 instead of 12.4 and 13.4, and the comma separated value for
`--exclude-group` it used triggers a PHPUnit deprecation which, together with
the setting `failOnPhpunitDeprecation`, fails the whole suite regardless of
the test results.

## Verification

Verified locally for both TYPO3 versions with PHP 8.2, after the rebase onto
current `main`:

* Static checks, PHPStan, PHPMD, `psrVerify`, `composerUnused`, the Composer
  and workflow linters, `shellcheck` and the documentation rendering pass.
* TYPO3 13.4: unit `OK (11 tests)`, functional `OK (72 tests, 106 assertions)`.
* TYPO3 14.3: unit `OK (11 tests)`, functional `OK (72 tests, 106 assertions)`.
* All four version/suite combinations of a mismatched run fail on the pinned
  test with the hint naming the version the run was made for:

  | installed | run with | unit | functional |
  | --- | --- | --- | --- |
  | 13.4 | `-t 14.3` | fails | fails |
  | 14.3 | `-t 13.4` | fails | fails |

Supersedes #2107.


Tests/VersionCompatibilityTestsTrait.php
```diff
@@ -0,0 +1,71 @@
+<?php
+
+declare(strict_types=1);
+
+namespace TTN\Tea\Tests;
```
oliverklee: Please let's move this to a dedicated directory/namespace for test helpers, e.g., `TTN\Tea\Tests\Support`.

Tests/Support/VersionCompatibilityTestsTrait.php
```diff
@@ -0,0 +1,71 @@
...
+ * recognize.
+ *
+ * `runTests.sh` passes `--exclude-group not-core-<version>` for the version
+ * given with `-t`, so exactly the pinned test of that version remains, and it
+ * fails right away if another version is installed.
+ *
+ * The group names are written out rather than composed from constants, so that
+ * they can be found together with the `--exclude-group` arguments which use
+ * them.
+ *
```
oliverklee, line 29: We should ensure that this trait only gets used for testcases using [`@require-extends`](https://phpstan.org/writing-php-code/phpdocs-basics#enforcing-class-inheritance-for-interfaces-and-traits):

`@require-extends BaseTestCase` or `@require-extends TestCase` (depending on how much we want to rely on/expose TYPO3 TF internals)

Build/Scripts/runTests.sh
```diff
@@ -284,11 +284,14 @@ Options:
...
             - 18    maintained until 2030-11-14
 
-    -t <13.4>
-        Only with -s composerUpdateMin|composerUpdateMax
+    -t <13.4|14.3>
+        Only with -s composerUpdateMin|composerUpdateMax|phpstan|phpstanGenerateBaseline|unit|unitRandom|functional
         Specifies the TYPO3 CORE Version to be used
             - 13.4: (default) use TYPO3 v13
             - 14.3: use TYPO3 v14
+        For the test suites this selects the tests which only apply to one TYPO3 version.
```
oliverklee:
```suggestion
        For the test suites, this selects the tests which only apply to one TYPO3 version.
```

Build/Scripts/runTests.sh
```diff
@@ -284,11 +284,14 @@ Options:
...
+        Use the version the dependencies have been installed for, a different one lets the
```
oliverklee:
```suggestion
        Use the version the dependencies have been installed for. A different one lets the
```

Tests/Functional/ExtensionLoadedTest.php
```diff
@@ -19,8 +19,12 @@ final class ExtensionLoadedTest extends FunctionalTestCase
     #[Test]
     public function isLoaded(): void
```
oliverklee:
```suggestion
    public function isLoadedWithExtensionKey(): void
```

oliverklee (changes requested):

Tests/VersionCompatibilityTestsTrait.php
sbuerk: done

Tests/Support/VersionCompatibilityTestsTrait.php
sbuerk: done

Build/Scripts/runTests.sh
sbuerk: done

Build/Scripts/runTests.sh
sbuerk: done

Tests/Functional/ExtensionLoadedTest.php
sbuerk: done

oliverklee (approved):

PR #2338, opened 2026-09-16
[TASK] Verification duplicate of #2336 (do not merge)

sbuerk:
**Verification only — please do not merge or review. This is a temporary
duplicate of #2336 and will be closed once it has served its purpose.**

Same branch content as #2336, pushed to a branch in this repository instead of
to a fork, to answer one question: does the Coveralls comment fail to appear
because the pull request comes from a fork, or because of the workflow change in
this series?

The comparison so far is confounded. #2336 (from a fork) runs the reworked
workflow and has had no Coveralls comment for hours. #2335 (from a branch in
this repository) runs the *old* workflow and got its comment 2.5 minutes after
its build. Those two differ in two ways at once, so neither explanation is
proven.

This pull request holds the reworked workflow *and* comes from a branch in this
repository. That isolates the variable:

* if a Coveralls comment appears here, the workflow change is not the cause and
  the fork origin is;
* if no comment appears here, the workflow change is the cause and #2336 is not
  the problem.

Background: #1802, and the real pull request #2336.


sbuerk:
Verification done, closing.

Result: this pull request — a branch in this repository, opened as a draft, with
the same commits and the same reworked workflow as #2336 — received its Coveralls
comment 66 seconds after the upload.

#2336, which is the same content from a fork, has had none for hours.

So the reworked coverage workflow is not the cause, and neither is the draft
state. The difference is that the pull request comes from a fork. Details and
the surrounding measurements are collected in the issue linked below.

The real pull request stays open: #2336.

sbuerk:
Verification round two done, closing again. The branch is kept this time so this
can be reopened without recreating it.

The coverage job now lives in the CI workflow and waits for the unit and
functional tests (`needs:` plus `if: ${{ !cancelled() }}`). That part works: in
the run on this head the last test job ended at 21:42:39Z and the coverage job
started at 21:42:44Z, and there is no separate "Code coverage" workflow run any
more.

It did not change the Coveralls behaviour, which is the point worth recording:

* here, from a branch in this repository, Coveralls **updated** its existing
  comment in place at 21:48:11Z, citing the new build;
* #2336, byte-identical commits and the same workflow but from a fork, still has
  no comment at all.

So deferring the coverage job until after the tests is not what governs the
comment. Fork origin remains the only variable that separates the two.

One unrelated note: `Functional tests (13.4, 8.2, sqlite, Max)` failed in this
run while pulling a container image — `connection reset by peer`, exit 125 — and
the same leg on the same commit passed in #2336's run. An infrastructure flake,
not a defect in the change.

Collected in #2339. The real pull request stays open: #2336.

